PR #26, opened 2024-03-17
feat(rgbpp-sdk/btc): btc rgbpp transactions

ShookLyngs:
## Main changes
1. Add sendUtxos() API for building BTC transactions with predefined inputs/outputs
2. Add sendRgbppUtxos() API for building RGBPP sync transactions on BTC
3. Rewrite the satoshi/fee collection process and improve flexibility
4. Rename types/constants/functions
    - constant: MIN_COLLECTABLE_SATOSHI -> BTC_UTXO_DUST_LIMIT
    - type: UnspentOutput -> Utxo
5. Remove `networkType` prop from BTC APIs

## Next
- [x] RGBPP/CKB transaction adapter API https://github.com/ckb-cell/rgbpp-sdk/pull/26/commits/bbc9a06a71af04cdb1e2b1768e96c77c8260ddbb
- [ ] #29 
- [ ] #32 

### The sendUtxos() API

```ts
const psbt = await sendUtxos({
  inputs: [
    {
      txid: 'txid',
      vout: 1,
      value: 546,
      address: 'btc_address',
      addressType: AddressType.P2WPKH,
      scriptPk: 'script_publickey_hex',
    },
  ],
  outputs: [
    {
      data: 'commentment_hex', // RGBPP commitment
      value: 0,
      fixed: true, // the output will not be changed or deducted
    },
    {
      address: 'to_address',
      value: RGBPP_UTXO_DUST_LIMIT, // the default RGBPP UTXO dust_limit is 546
      minUtxoSatoshi: RGBPP_UTXO_DUST_LIMIT, // customize the dust_limit of the output
      protected: true, // free satoshi can be deducted, but the output cannot be removed (freeSatoshi = value - minUtxoSatoshi)
    },
  ],
  from: account.address,
  source,
})
```

### The sendRgbppUtxos() API

```ts
const ckbVirtualTx: RawTransaction = {} as any;
const ckbCollector = new Collector({
  ckbNodeUrl: 'ckb_node_url', 
  ckbIndexerUrl: 'ckb_indexer_url'
});
const psbt = await sendRgbppUtxos({
  ckbVirtualTx, // a CKB virtual tx containing L1 -> L1 or L1 -> L2 action
  paymaster: { // if paymaster cell was included in the ckbVirtualTx, pay to paymaster
    address: 'paymaster_btc_address',
    value: 10000,
  },
  commitment: 'rgbpp_tx_commitment',
  tos: [ // the address of the generating outputs, optional, default is "from"
    'transfer_rgbpp_to_btc_address',
  ],

  ckbCollector,
  rgbppMinUtxoSatoshi: RGBPP_UTXO_DUST_LIMIT, // optional, default to 546

  from: accounts.address,
  fromPubkey: account.publicKey, // if "from" is a P2TR address, "fromPubkey" is required 
  changeAddress: 'address_to_return_change', // optional, where should the change satoshi be returned to
  minUtxoSatoshi: BTC_UTXO_DUST_LIMIT, // optional, default to 1000, officially should be 1,0000
  source,
});
```

Here's a L1 -> L1 example transaction, showing how A transfers 400 RGBPP to B:
```ts
// CKB
inputs:
- RGBPP_XUDT_1, lock=A, amount=1000
- CELL, lock=A
outputs:
- RGBPP_XUDT_2, lock=B, amount=400
- RGBPP_XUDT_3, lock=A, amount=600
- CELL, lock=A

// BTC
ins:
- RGBPP_UTXO_1, address=A
- UTXO, address=A
outs:
- OP_RETURN_UTXO, data=commitment
- RGBPP_UTXO_2, address=B
- RGBPP_UTXO_3, address=A
- UTXO, address=A
```

Since A is transferring some RGBPP assets to B, when building the BtcVirtualTx, you should specify the `address` for `RGBPP_UTXO_2`, otherwise by default the address of the RGBPP_UTXO_2 will be set to A (as the `from` prop specified):
```ts
tos: {
  'RGBPP_UTXO_2_address', // The address of the outs[1], optional, default is "from"
  'RGBPP_UTXO_3_address', // The address of the outs[2], optional, default is "from"
}
```

### Output markers

This output with the `fixed` mark will NOT be deducted for satoshi/fee collection:
```ts
{
  address: 'to_address',
  value: 10000,
  fixed: true,
}
```

This output is marked as `protected`, so the free satoshi in this output can be deducted (but this output cannot be removed):
```ts
{
  address: 'to_address',
  value: 10000,
  protected: true,
}
```

This output without any marker can be deducted during the satoshi/fee collection process, and this output can be removed from the transaction if necessary:
```ts
{
  address: 'to_address',
  value: 10000,
}
```

### Satoshi/fee collection

The satoshi/fee collection process has been rewritten for comlex situations like in a RGBPP transaction:
1. If there are outputs matching the `(output.protected == true && output.fixed == false)` condition, the collection process will try to collect free satoshi from those outputs. An output's free satoshi is `output.value - output.minUtxoSatoshi`.
2. If the sum of satoshi in the transaction matches `(sum(inputs) > sum(outputs))`, then the change satoshi can be returned in the first output that matches the `(output.fixed == false && output.address == changeAddress)` condition.

BTW, since the entire collection process has been rewritten, I believe we need to conduct thorough tests on the fee estimation part. Although I have added many tests to cover various scenarios, there may still be additional issues that need to be discovered. Therefore, we could greatly benefit from more tests and reviews. cc @Dawn-github.


duanyytop:
I found that the constructor of `DataSource` needs to pass the `NetworkType`, so why `sendUtxos` and `sendRgbppUtxos` also need to pass the `NetworkType`?

ShookLyngs:
> I found that the constructor of `DataSource` needs to pass the `NetworkType`, so why `sendUtxos` and `sendRgbppUtxos` also need to pass the `NetworkType`?

You're right, the `networkType` prop can be passed in the DataSource instance. 
It has been removed from the props of the APIs (sendBtc/sendUtxos/sendRgbppUtxos).

packages/btc/src/transaction/build.ts
```diff
@@ -42,110 +46,247 @@ export class TxBuilder {
...
+    }
+
+    if (amount < this.minUtxoSatoshi) {
+      const { collected } = await this.injectSatoshi({
         address,
-        pubkey,
-        fee: estimatedFee,
-        extraChange: nextExtraChange,
+        publicKey,
+        targetAmount: amount,
```
Flouse, line 279:
```suggestion
        targetAmount: this.minUtxoSatoshi - amount,
```

Flouse (changes requested):

packages/btc/src/query/source.ts
```diff
@@ -23,7 +43,7 @@ export class DataSource {
         const bOrder = `${b.status.block_height}${b.vout}`;
         return Number(aOrder) - Number(bOrder);
       })
-      .map((row): UnspentOutput => {
+      .map((row): Utxo => {
         return {
```
Flouse, line 47: let's say:
- a.status.block_height = 100
- b.status.block_height = 20
- a.vout = b.vout = 0

"1000" < "200" === true

packages/btc/src/api/sendRgbppUtxos.ts
```diff
@@ -0,0 +1,169 @@
...
+
+      // If input.type !== null，update lastTypeInput
+      lastTypeInputIndex = i;
+    }
+
+    // If input.lock == RgbppLock, add to inputs if:
+    // 1. input.lock.args can be unpacked to RgbppLockArgs
+    // 2. utxo can be found via the DataSource.getUtxo() API
+    // 3. utxo.scriptPk == addressToScriptPk(props.from)
+    if (isRgbppLock || isRgbppTimeLock) {
```
Flouse, line 69:
```suggestion
    if (isRgbppLock) {
```
no need to check `isRgbppTimeLock` in inputs

Flouse (commented):
Other codes LGTM.

packages/btc/src/api/sendRgbppUtxos.ts
```diff
@@ -0,0 +1,169 @@
...
+  // Handle and check inputs
+  const inputCells = ckbTxSkeleton.get('inputs');
+  for (let i = 0; i < inputCells.size; i++) {
+    const input = inputCells.get(i)!;
+    const isRgbppLock = input.cellOutput.lock.codeHash === props.rgbppLockCodeHash;
+    const isRgbppTimeLock = input.cellOutput.lock.codeHash === props.rgbppTimeLockCodeHash;
+
+    // If input.type !== null, input.lock must be RgbppLock or RgbppTimeLock
+    if (input.cellOutput.type) {
+      if (!isRgbppLock && !isRgbppTimeLock) {
```
Flouse, line 57:
```suggestion
      if (!isRgbppLock) {
```

duanyytop:
```
  ckbNodeUrl: 'https://testnet.ckb.dev/rpc',
  rgbppLockCodeHash: 'rgbpp_lock_code_hash',
  rgbppTimeLockCodeHash: 'rgbpp_time_lock_code_hash',
```
I suggest that we should remove the parameters and the commitment calculation has been done in CKB package

packages/btc/src/api/sendRgbppUtxos.ts
```diff
@@ -0,0 +1,169 @@
...
+export async function sendRgbppUtxos(props: {
+  ckbVirtualTx: RawTransaction;
+  paymaster: TxAddressOutput;
+  commitment: Hash;
+  tos?: string[];
+
+  ckbNodeUrl: string;
+  rgbppLockCodeHash: Hash;
+  rgbppTimeLockCodeHash: Hash;
+  rgbppMinUtxoSatoshi?: number;
```
duanyytop, line 22:
```
  ckbNodeUrl: string;
  rgbppLockCodeHash: Hash;
  rgbppTimeLockCodeHash: Hash;
```
I suggest that we should remove the parameters and the commitment calculation has been done in CKB package

packages/btc/src/query/source.ts
ShookLyngs: Those stirngs will be converted into numbers before comparing, so actually it's like this: `Number('1000') < Number('200') === false`. But you reminded me that there will be issues if the digit difference between `a.vout` and `b.vout`, like: "100|0" vs "20|20".

packages/btc/src/api/sendRgbppUtxos.ts
ShookLyngs: 1. Yes, I believe the commitment validation logic can be refactored using the `@rgbpp-sdk/ckb` lib 👍. 
2. However, the props you mentioned may be impossible to remove due to the needs of transaction construction. These props are used to query `Cell` from `OutPoint` in the inputs, and I have to check which input is an `RgbppLock` cell and which output is an `RgbppTimeLock` cell.

duanyytop: OK, I see. 

But you can get the RGBPP lock script and BTC time lock script information from the CKB package(src/constants). You can use the Collector of CKB package to query cells.

I think the parameters are not necessary for developers and they make the function usage more complicated.

packages/btc/src/ckb/commitment.ts
```diff
@@ -0,0 +1,60 @@
+import { sha256 } from 'js-sha256';
```
duanyytop, line 1: I think the `ckb` module which includes `commitment.ts`, `molecule.ts`, and `rpc.ts` should be removed from the BTC package. Any method about CKB should be implemented in CKB package and is called by BTC package. 

If the CKB package has no related methods, I will add them ASAP.

ShookLyngs: Okay, for that I think I need the following APIs: 
- calculateCommitment: calculate commitment from RawTransaction (provided)
- isRgbppLockCell: check if an input (or cell, see what you can do to simplify things) is an RgbppLock cell
- isRgbppTimeLockCell: similar logic as `isRgbppLockCell`, checks if an input/cell is an RgbppTimeLock cell
- getCellByOutPoint: Query a cell by an OutPoint, do the querying work, but if the above methods accepts `Input` instead of `Cell`, this method is redundant

duanyytop: Added in https://github.com/ckb-cell/rgbpp-sdk/pull/34

ShookLyngs: Refactored, removed most ckb deps, leaving a few types from lumos: https://github.com/ckb-cell/rgbpp-sdk/pull/26/commits/e630f922b208bae3edc63310404eb6714759a8ce

ShookLyngs: Replaced lock checking with cell output checking: https://github.com/ckb-cell/rgbpp-sdk/pull/26/commits/b93177be98021874bb7be6c4a74d3813fafe2954

duanyytop (approved):

Flouse (approved):

PR #56, opened 2024-03-26
refactor!(rgbpp-sdk/btc): split BtcAssetsApi as a separate package

ShookLyngs:
## Changes
1. BREAKING CHANGE: Split the code/docs of BtcAssetsApi as a separate package
   - The class is now exported from `@rgbpp-sdk/service` instead of `@rgbpp-sdk/btc`
   - BTC APIs in the BtcAssetsApi are now renamed to be differentiated from RGBPP APIs
2. Import BtcAssetsApi from `@rgbpp-sdk/service` in the `@rgbpp-sdk/btc`


packages/service/src/service/base.ts
```diff
@@ -0,0 +1,115 @@
...
+    try {
+      text = await res.text();
+      json = JSON.parse(text);
+      ok = json?.ok ?? res.ok ?? false;
+    } catch {
+      // do nothing
+    }
+
+    if (!json) {
+      if (status === 200) {
```
ahonn, line 61: Recommended to judge the status code after the response is returned, rather than after JSON parse

ShookLyngs: Moved it before the JSON parse, and blocks it if status !== 200, do you think it's better?

duanyytop (approved):

ShookLyngs:
Rebased onto the main branch to resolve conflicts.

Flouse (approved):

PR #70, opened 2024-03-29
feat(rgbpp-sdk/btc): provide default configs for testnet/mainnet

ShookLyngs:
## Main changes
1. Provide default configs (including dust limit for BTC/RGBPP UTXOs and more) for testnet and mainnet
2. Add actual dust limit validation in the `TxBuilder.addOutput()` method to prevent unexpected issues
3. Refactor P2WPKH transfer tests to a single recipient test, simplifying things
4. Set the default timeout of tests to 10 sec (suggested by @duanyytop)

### Dust limit on `MAINNET`

The dust limit of UTXOs on the mainnet should be set to `1,0000` instead of `1000` in the testnet. 

The PR differenciated the testnet/mainnet env and now provides configs for different network. When you're using the SDK on mainnet, you should transfer at least `1,0000` satoshi to the recipient by default. For example, If you try to define an output on the mainnet and its value < minUtxoSatoshi, the TxBuidler throws:

```
Error: Output defined value is below the dust limit: expected 10000, but defined 1000
```

### Dust limit for RGBPP UTXO

The dust limit of RGBPP UTXO on both mainnet/testnet are the same: `546` satoshi.

But how can the SDK know which output is an RGBPP UTXO? Well, if you use the `sendRgbppUtxos` API, then the SDK handles it by default. If you tend to not use the API, and you want to build your own RGBPP transctions with the `sendBtc` or `sendUtxos` API, you should specify `minUtxoSatoshi` for each RGBPP output manually.

```ts
sendUtxos({
  ...,
  outputs: {
    address: 'bc1q...',
    value: 546,
    minUtxoSatoshi: 546,
  },
})
```

### Dust limit for OP_RETURN

The dust limit for OP_RETURN UTXOs is an exception, the SDK doesn't validte its value.
In our own service, an OP_RETURN UTXO should contain 0 value by default.

```ts
sendUtxos({
  ...,
  outputs: {
    data: '0001',
    value: 0,
  },
})
```


packages/btc/src/preset/network.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+export function networkToNetworkType(network: bitcoin.Network): NetworkType {
+  const config = networkToConfig(network);
+  return config.networkType;
+}
+
+/**
+ * The rgbpp-sdk/ckb accepts a "isMainnet" property to indicate the network type.
+ */
+export function networkTypeToIsCkbMainnet(networkType: NetworkType): boolean {
```
duanyytop, line 24: The name of the `networkTypeToIsCkbMainnet` is a bit strange. Just make it simple and call it `checkIsCkbMainnet`.

packages/btc/src/transaction/build.ts
```diff
@@ -298,11 +313,13 @@ export class TxBuilder {
...
 
-  async calculateFee(addressType: AddressType): Promise<number> {
-    const feeRate = this.feeRate ?? this.defaultFeeRate;
-    if (!feeRate) {
+  async calculateFee(addressType: AddressType, feeRate?: number): Promise<number> {
+    if (!feeRate && !this.feeRate) {
       throw new TxBuildError(ErrorCodes.INVALID_FEE_RATE, `${ErrorMessages[ErrorCodes.INVALID_FEE_RATE]}: ${feeRate}`);
     }
+    if (!feeRate) {
+      feeRate = this.feeRate!;
```
duanyytop, line 321: Changing function parameters directly is not good coding practice. You can use a new variable like this: 
```typescript
const txFeeRate = feeRate ?? this.feeRate!
```

duanyytop (approved):

Flouse (changes requested):
I think `packages/btc/README.md` should be updated too.

Flouse (approved):

PR #90, opened 2024-04-01
fix(rgbpp-sdk/btc): fee estimation/collection issues

ShookLyngs:
## Changes
1. Fix when `sum(ins) > sum(outs) && change < fee`, the returned change is incorrect (found by @duanyytop)
2. Fix when `outs` has only an `OP_RETURN` output, and `ins` has no UTXOs specified, the collection is skipped

### When `sum(ins) > sum(outs) && change < fee`

Previously when the sum of ins is greater than the sum of outs, the process goes in the returning phase. But when calculating the return amount with `extraSatoshiInInputs - fee`, the `fee` can actually be greater then the `extraSatoshiInInputs`, in that case, the returning amount is a negative number.

To address the issue, I changed the check statement:
- from: `if (sum(ins) - sum(outs) > 0) { return change } else { colelct satoshi }`
- to: `if (sum(ins) - sum(outs) - fee > 0) { return change } else { colelct satoshi }`

The logic change should prevent errors when there is change to be returned, but the change is less than the fee. In this situation, it should always collect for more instead of returning a negative value.

### When `sum(ins) == sum(outs) == 0`

When the BTC TX initialized as the following, the `sum(ins)` and `sum(outs)` are both zero:
```yaml
ins(0):

outs(1):
- OP_RETURN output, value=0
```

When the inputs list has no length, the PSBT cannot be signed or finalized, and the fee estimation cannot be processed if the PSBT isn't finalized. The process will throw an error in that case, even though this is a possible situation for an initial transaction.

To resolve this issue, I added a protection check:
```ts
const safeToProcess = inputsTotal > 0 || previousFee > 0;
if (safeToProcess && canReturnChange) { 
  // ... return change
} else {
  const protectionAmount = !safeToProcess ? 1 : 0;
  await collectSatoshi({ amount: targetAmount + protectionAmount, ... })
}
``` 

The protection ensures that if the fee is 0 (not calculated yet) and the inputs list is empty, the collection must be processed once with the targeet of at least 1 satoshi, in order for the fee estimation to be calculated later on.

### DIFF
--- a/packages/btc/src/transaction/build.ts
+++ b/packages/btc/src/transaction/build.ts
@@ -137,9 +137,10 @@ export class TxBuilder {
     let currentFee: number = 0;
     let previousFee: number = 0;
     while (true) {
-      const { needCollect, needReturn } = this.summary();
+      const { needCollect, needReturn, inputsTotal } = this.summary();
+      const safeToProcess = inputsTotal > 0 || previousFee > 0;
       const returnAmount = needReturn - previousFee;
-      if (returnAmount > 0) {
+      if (safeToProcess && returnAmount > 0) {
         // If sum(inputs) - sum(outputs) > fee, return change while deducting fee
         // Note, should not deduct fee from outputs while also returning change at the same time
         await this.injectChange({
@@ -149,7 +150,8 @@ export class TxBuilder {
           fromPublicKey: publicKey,
         });
       } else {
-        const targetAmount = needCollect + previousFee - needReturn;
+        const protectionAmount = !safeToProcess ? 1 : 0;
+        const targetAmount = needCollect - needReturn + previousFee + protectionAmount;
         await this.injectSatoshi({
           address,
           publicKey,
--- a/packages/btc/src/transaction/build.ts
+++ b/packages/btc/src/transaction/build.ts
@@ -150,7 +150,7 @@ export class TxBuilder {
           fromPublicKey: publicKey,
         });
       } else {
-        const protectionAmount = !safeToProcess ? 1 : 0;
+        const protectionAmount = safeToProcess ? 0 : 1;
         const targetAmount = needCollect - needReturn + previousFee + protectionAmount;
         await this.injectSatoshi({
           address,
@@ -383,16 +383,19 @@ export class TxBuilder {
   }
 
   summary() {
-    const sumOfInputs = this.inputs.reduce((acc, input) => acc + input.utxo.value, 0);
-    const sumOfOutputs = this.outputs.reduce((acc, output) => acc + output.value, 0);
+    const inputsTotal = this.inputs.reduce((acc, input) => acc + input.utxo.value, 0);
+    const outputsTotal = this.outputs.reduce((acc, output) => acc + output.value, 0);
+
+    const inputsRemaining = inputsTotal - outputsTotal;
+    const outputsRemaining = outputsTotal - inputsTotal;
 
     return {
-      inputsTotal: sumOfInputs,
-      outputsTotal: sumOfOutputs,
-      inputsRemaining: sumOfInputs - sumOfOutputs,
-      outputsRemaining: sumOfOutputs - sumOfInputs,
-      needReturn: sumOfInputs > sumOfOutputs ? sumOfInputs - sumOfOutputs : 0,
-      needCollect: sumOfOutputs > sumOfInputs ? sumOfOutputs - sumOfInputs : 0,
+      inputsTotal,
+      outputsTotal,
+      inputsRemaining,
+      outputsRemaining,
+      needReturn: inputsRemaining > 0 ? inputsRemaining : 0,
+      needCollect: outputsRemaining > 0 ? outputsRemaining : 0,
     };
   }
 
--- a/packages/btc/src/transaction/build.ts
+++ b/packages/btc/src/transaction/build.ts
@@ -1,11 +1,11 @@
 import clone from 'lodash/cloneDeep';
 import { bitcoin } from '../bitcoin';
 import { DataSource } from '../query/source';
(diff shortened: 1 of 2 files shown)


packages/btc/src/transaction/build.ts
```diff
@@ -137,18 +137,21 @@ export class TxBuilder {
...
         await this.injectChange({
           address: changeAddress ?? address,
           amount: returnAmount,
-          publicKey,
+          fromAddress: address,
+          fromPublicKey: publicKey,
         });
       } else {
-        const targetAmount = inputsNeeding + previousFee;
+        const protectionAmount = !safeToProcess ? 1 : 0;
```
duanyytop, line 153:
```suggestion
        const protectionAmount = safeToProcess ? 0 : 1;
```

packages/btc/tests/Transaction.test.ts
```diff
@@ -266,7 +296,7 @@ describe('Transaction', () => {
       // const res = await service.sendBtcTransaction(tx.toHex());
       // console.log(`explorer: https://mempool.space/testnet/tx/${res.txid}`);
     });
-    it('Transfer fixed UTXO, sum(ins) > sum(outs), but no collection', async () => {
+    it('Transfer fixed UTXO, sum(ins) > sum(outs), change > fee, change > fee + minUtxoSatoshi', async () => {
```
duanyytop, line 299:
```suggestion
    it('Transfer fixed UTXO, sum(ins) > sum(outs), change > fee + minUtxoSatoshi', async () => {
```

packages/btc/tests/Transaction.test.ts
```diff
@@ -304,7 +334,86 @@ describe('Transaction', () => {
       // const res = await service.sendBtcTransaction(tx.toHex());
       // console.log(`explorer: https://mempool.space/testnet/tx/${res.txid}`);
     });
-    it('Transfer fixed UTXO, sum(ins) > sum(outs), the fee is prepaid', async () => {
+    it('Transfer fixed UTXO, sum(ins) > sum(outs), change > fee, change = fee + minUtxoSatoshi', async () => {
```
duanyytop, line 337:
```suggestion
    it('Transfer fixed UTXO, sum(ins) > sum(outs), change = fee + minUtxoSatoshi', async () => {
```

packages/btc/tests/Transaction.test.ts
```diff
@@ -471,6 +580,47 @@ describe('Transaction', () => {
...
+        source,
+      });
+
+      // Sign & finalize inputs
+      const psbt = builder.toPsbt();
+      psbt.signAllInputs(accounts.charlie.keyPair);
+      psbt.finalizeAllInputs();
+
+      expect(psbt.txInputs.length).toBeGreaterThanOrEqual(2);
+      expect(psbt.txOutputs).toHaveLength(1);
```
duanyytop, line 613: If there are two or more inputs, the change output is required?

duanyytop, line 613: If there are two or more inputs, is the change output required?

packages/btc/src/transaction/build.ts
```diff
@@ -381,11 +388,11 @@ export class TxBuilder {
...
       inputsTotal: sumOfInputs,
-      inputsRemaining: sumOfInputs - sumOfOutputs,
-      inputsNeeding: sumOfOutputs > sumOfInputs ? sumOfOutputs - sumOfInputs : 0,
       outputsTotal: sumOfOutputs,
+      inputsRemaining: sumOfInputs - sumOfOutputs,
       outputsRemaining: sumOfOutputs - sumOfInputs,
-      outputsNeeding: sumOfInputs > sumOfOutputs ? sumOfInputs - sumOfOutputs : 0,
+      needReturn: sumOfInputs > sumOfOutputs ? sumOfInputs - sumOfOutputs : 0,
+      needCollect: sumOfOutputs > sumOfInputs ? sumOfOutputs - sumOfInputs : 0,
     };
```
ahonn, line 396: I prefer to get `needReturn` and `needCollect` based on `inputsRemaining`.

```typescript
const inputsRemaining = sumOfInputs - sumOfOutputs;
const needReturn = inputsRemaining > 0 ? inputsRemainning : 0;
const needCollect = inputsRemaining < 0 ? -inputsRemaining : 0;
```

I think this is clearer, expressing that the values ​​​​of `needReturn` and `needCollect` depend on how many inputs remain.

packages/btc/tests/Transaction.test.ts
ShookLyngs: In this test, the change is designed to be returned in outs[0] because it is a `protected output`. A protected output cannot be removed from the outs, but the free value (free value = output.value - minUtxoSatoshi) defined in the output can be deducted or added.

packages/btc/src/transaction/build.ts
ahonn: We can even just calculate `sumOfInputs - sumOfOutputs`, `outputsRemaining` being the negative of the result of this calculation.

packages/btc/tests/Transaction.test.ts
duanyytop: Make sense

packages/btc/src/transaction/build.ts
ShookLyngs: Yes, when I looked back in the code, I also felt that it was a bit hard to understand, and this is also the reason I changed the naming of the variables here.

duanyytop (approved):

Flouse (approved):

ahonn (approved):

Dawn-githup (approved):

PR #91, opened 2024-04-01
feat(btc|service): add option to filter confirmed utxos and validate inputs

### LINKED ISSUE #87: Add a new optional parameter `transactionConfirmed` to the `sendBtc` method
An app requests a feature in the `sendBtc` method of @rgbpp-sdk/btc, to prevent consecutive transactions.

## Add a new optional parameter `transactionConfirmed` to the sendBtc method
The purpose of this parameter is to **prevent consecutive transactions** from being sent before the previous one is confirmed.

### Proposed Changes
1. Set the default value of `transactionConfirmed` to `false`.
2. Implement logic to check the value of `transactionConfirmed`:
   - If `transactionConfirmed` is `true`, proceed with the transaction only if the previous transactions of the inputs has been confirmed.
     More specifically, the `input.UTXO.status.confirmed` should be true in this case:
     https://github.com/ckb-cell/rgbpp-sdk/blob/1458e40291b97ebfbbfc2007df83a24043e593a5/packages/service/src/types/btc.ts#L65-L75
   - If `transactionConfirmed` is `false`, proceed with the transaction regardless of the status of the previous transaction.
3. If the previous transaction has not been confirmed and `transactionConfirmed` is `true`, throw an appropriate error.

### Example Usage
```typescript
const psbt = await sendBtc({
  from: address,
  tos: [
    {
      address: TARGET_ADDRESS,
      value: 546,
    },
  ],
  transactionConfirmed: true, // If the previous transaction is not confirmed, an error will be thrown
  fromPubkey: pubkey,
  changeAddress: address,
  feeRate: 1,
  source,
});
```

### Benefits
Provides more control and flexibility for developers

ShookLyngs:
## Changes
1. Support the "only_confirmed" param in the BtcAssetsApi.getUtxos() API (refer to [btc-assets-api#40](https://github.com/ckb-cell/btc-assets-api/pull/40) and [btc-assets-api#41](https://github.com/ckb-cell/btc-assets-api/pull/41))
2. Support the "onlyConfirmedUtxos" option in the BTC APIs (`sendBtc`, `sendUtxos`, `sendRgbppUtxos`, and the BTC Builder APIs), and also validate inputs to check if each UTXO is confirmed

### Option `onlyConfirmedUtxos` in BTC APIs

This is a `boolean` option that tells the API to look for confirmed UTXOs when collecting inputs for construction and fee. If not enough confirmed UTXOs to construct the transaction, the API will throw an INSUFFICIENT_UTXO error.

```ts
sendBtc({
  // ...
  onlyConfirmedUtxos: true, // optional, default to false
});
```

## Discussion

- The name of  the "onlyConfirmedUtxos" option is inherited from the naming of btc-assets-api (refer to [btc-assets-api#40](https://github.com/ckb-cell/btc-assets-api/pull/40) and [btc-assets-api#41](https://github.com/ckb-cell/btc-assets-api/pull/41)), which is different than the [feature requested](https://github.com/ckb-cell/rgbpp-sdk/issues/87).
- Do you prefer "transactionConfirmed" from the feature request as a better name for the option? Do you think "onlyConfirmedUtxos" is fine for now? Or do you have better naming suggestions? (validateInputs? filterConfirmedUtxos? requireConfirmedInputs? requireConfirmedUtxos? requireConfirmedInputs?)

## Related issues
- Resolves #87

### DIFF
--- a/packages/btc/src/query/source.ts
+++ b/packages/btc/src/query/source.ts
@@ -20,9 +20,10 @@ export class DataSource {
   }
 
   // Query a UTXO from the service.
-  // Will throw error if the target output is unspendable
-  async getUtxo(hash: string, index: number): Promise<Utxo | undefined> {
-    const output = await this.getOutput(hash, index);
+  // Will throw error if the target output is unspendable.
+  // When set "confirmed = true", will throw error if the output is unconfirmed.
+  async getUtxo(hash: string, index: number, requireConfirmed?: boolean): Promise<Utxo | undefined> {
+    const output = await this.getOutput(hash, index, requireConfirmed);
     if (output && !('address' in output)) {
       throw TxBuildError.withComment(ErrorCodes.UNSPENDABLE_OUTPUT, `hash: ${hash}, index: ${index}`);
     }
@@ -32,11 +33,15 @@ export class DataSource {
 
   // Query an output from the service.
   // Both unspent or unspendable output can be queried from the API.
-  async getOutput(hash: string, index: number): Promise<Output | Utxo | undefined> {
+  // When set "confirmed = true", will throw error if the output is unconfirmed.
+  async getOutput(hash: string, index: number, requireConfirmed?: boolean): Promise<Output | Utxo | undefined> {
     const txId = remove0x(hash);
     const tx = await this.service.getBtcTransaction(txId);
     if (!tx) {
       return void 0;
+    }
+    if (requireConfirmed && !tx.status.confirmed) {
+      throw TxBuildError.withComment(ErrorCodes.UNCONFIRMED_UTXO, `hash: ${hash}, index: ${index}`);
     }
     const vout = tx.vout[index];
     if (!vout) {
@@ -61,6 +66,11 @@ export class DataSource {
       address: vout.scriptpubkey_address,
       addressType: getAddressType(vout.scriptpubkey_address),
     } as Utxo;
+  }
+
+  async isTransactionConfirmed(hash: string): Promise<boolean> {
+    const tx = await this.service.getBtcTransaction(remove0x(hash));
+    return tx.status.confirmed;
   }
 
   async getUtxos(address: string, params?: BtcApiUtxoParams): Promise<Utxo[]> {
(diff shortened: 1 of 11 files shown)


Flouse:
> * Do you prefer "transactionConfirmed" from the feature request as a better name for the option? Do you think "onlyConfirmedUtxos" is fine for now? Or do you have better naming suggestions? (validateInputs? filterConfirmedUtxos? requireConfirmedInputs? requireConfirmedUtxos? requireConfirmedInputs?)

`onlyConfirmedUtxos` is good.

packages/service/src/types/btc.ts
```diff
@@ -60,6 +60,7 @@ export interface BtcApiBalance {
 }
 
 export interface BtcApiUtxoParams {
+  only_confirmed?: boolean;
```
Flouse, line 63: Please update README too.
https://github.com/ckb-cell/rgbpp-sdk/blob/cb98e5c117e9d363624ca49329c9cd98163e87fd/packages/service/README.md#L184-L186

packages/btc/src/query/source.ts
```diff
@@ -63,6 +68,11 @@ export class DataSource {
     } as Utxo;
   }
 
+  async isTransactionConfirmed(hash: string): Promise<boolean> {
+    const tx = await this.service.getBtcTransaction(remove0x(hash));
```
Flouse, line 72: Sometimes a 404 error (ASSETS_API_RESOURCE_NOT_FOUND) can be thrown here without sufficient error context such as txid

image

Flouse (approved):

packages/btc/src/query/source.ts
ShookLyngs: Any suggestions on how to provide clearer context? A simple solution is that we can inject the request context into the error, which is shown in the following code. However, normally this method doesn't provide great TypeScript support.
```ts
try { 
  ... 
} catch (e: any) {
  if (e instanceof BtcAssetsApiError) {
    e.context = { txid };
  }
} 
```

Flouse: LGTM.
With a txid, the develop using SDK may be clearer to track the error.

Feel free to implement in another PR, maybe other errors need this kind of context too.

PR #127, opened 2024-04-22
feat: fill paymaster utxo output automatically

ahonn:
## Changes
Automatically obtain the paymaster info from the Assets API, and add the paymaster UTXO to the output of the BTC transaction.

- Assets API sets paymaster info
   - Verify when `sendRgbppUtxos` has a paymaster parameter. If inconsistent, an error will be thrown.
- Assets API does not set paymaster info
   - When `sendRgbppUtxos` has a paymaster parameter, it is used directly without verification.
   - No processing is done when `sendRgbppUtxos` does not pass the paymaster parameter.

Assets API is compatible with previous versions when paymaster info is not set.

## Reviews
@ShookLyngs @Flouse


packages/btc/src/query/source.ts
```diff
@@ -165,4 +166,19 @@ export class DataSource {
     const fees = await this.getRecommendedFeeRates();
     return fees.halfHourFee;
   }
+
+  async getPaymasterUTXOOuput(): Promise<TxAddressOutput | undefined> {
```
ShookLyngs, line 170:
```suggestion
  async getPaymasterOuput(): Promise<TxAddressOutput | undefined> {
```
The concept of `utxo` and `output` is redundant, maybe just keep the `output`?

packages/btc/src/api/sendRgbppUtxos.ts
```diff
@@ -164,6 +142,38 @@ export async function sendRgbppUtxosBuilder(props: SendRgbppUtxosProps): Promise
...
+    data: props.commitment,
+    fixed: true,
+    value: 0,
+  });
+
+  // Add outputs
+  merged.push(...btcOutputs);
+
+  // Add paymaster UTXO
+  const paymasterUTXOOutput = await props.source.getPaymasterUTXOOuput();
```
ShookLyngs, line 159:
```suggestion
  const paymasterOutput = await props.source.getPaymasterOuput();
```

ShookLyngs (approved):

Flouse (approved):

PR #130, opened 2024-04-23
feat(ckb | examples): Support Leap spores and add examples

duanyytop:
## Main Changes

- Supports spores leap operation, including leaping from BTC to CKB and leaping from CKB to BTC
- Add spore leaping examples, including leaping from BTC to CKB and leaping from CKB to BTC
- Add spore BTC time cells unlocked example
- Supports cobuild witness generation of spores batch transfers
- Add RGB++ spore document to README


ahonn (approved):

ShookLyngs (approved):

packages/ckb/src/rgbpp/btc-time.ts
```diff
@@ -128,13 +128,15 @@ export const signBtcTimeCellSpentTx = async ({
...
+  let emptyCells = await collector.getCells({
     lock: masterLock,
+    isDataEmpty: true,
     outputCapacityRange,
   });
   if (!emptyCells || emptyCells.length === 0) {
     throw new Error('No empty cell found');
   }
+  emptyCells = emptyCells.filter((cell) => !cell.output.type);
   const emptyInput: CKBComponents.CellInput = {
```
Flouse, line 140: code search: https://github.com/search?q=repo%3Ackb-cell%2Frgbpp-sdk+emptyCells+%3D+await+collector.getCells&type=code

other similar code with `emptyCells = await collector.getCells` may need to be updated too?
- isDataEmpty: true
- emptyCells = emptyCells.filter((cell) => !cell.output.type)

Flouse (changes requested):

packages/ckb/src/spore/leap.ts
```diff
@@ -0,0 +1,270 @@
...
+  toCkbAddress,
+  isMainnet,
+  witnessLockPlaceholderSize,
+  ckbFeeRate,
+}: LeapSporeFromBtcToCkbVirtualTxParams): Promise<SporeLeapVirtualTxResult> => {
+  const sporeRgbppLock = {
+    ...getRgbppLockScript(isMainnet),
+    args: append0x(sporeRgbppLockArgs),
+  };
+  const sporeCells = await collector.getCells({ lock: sporeRgbppLock, isDataEmpty: false });
```
Flouse, line 63:
```suggestion
  const sporeCells = await collector.getCells({ lock: sporeRgbppLock });
```

Should search all cells locked by the expected rgbpp-lock, including empty data cells.

packages/ckb/src/spore/leap.ts
```diff
@@ -0,0 +1,270 @@
...
+  if (!sporeCells || sporeCells.length === 0) {
+    throw new NoRgbppLiveCellError('No spore rgbpp cells found with the spore rgbpp lock args');
+  }
+  if (sporeCells.length > 1) {
+    throw new RgbppUtxoBindMultiSporesError('The UTXO is bound to multiple spores');
```
Flouse, line 68:
```suggestion
    throw new RgbppUtxoBindMultiSporesError('The BTC UTXO is bound to multiple CKB cells');
```

Flouse (changes requested):

packages/ckb/src/rgbpp/btc-time.ts
duanyytop: Under normal circumstances, cell data is not empty, meaning the cell type script is not null and vice versa is also true.

The `isDataEmpty` here defaults to true, so there is no problem keeping the original code in most cases.

I once accidentally created a cell during the debug phase. Its cell data was empty, but its cell type was not null. Even if the type is set to null during filtering, the returned cell list will still include this cell.

So I added a layer of code to filter type script here. To be on the safe side, this filtering should be added elsewhere.

packages/ckb/src/spore/leap.ts
duanyytop: Does Spore protocol allow cell data to be empty? If allowed, it does need to be added

Flouse: I think it's possible that an empty data cell could be bound to the same rgbpp-lock with spore a spore cell.

### All the rgbpp asset cells that are isomorphic with a specific BTC_UTXO, should be unlocked in one single transaction.

packages/ckb/src/rgbpp/btc-time.ts
Flouse: I see, `isDataEmpty` is default to true.

https://github.com/ckb-cell/rgbpp-sdk/blob/dbb81e36f61fae9a8eeb64d3fd022d7b051cf056/packages/ckb/src/collector/index.ts#L32

duanyytop:
I updated all the comments and code in https://github.com/ckb-cell/rgbpp-sdk/pull/130/commits/36613d1cc1c5125cf3c1c7c055c523b3e3681521 
@Flouse

packages/ckb/src/spore/leap.ts
duanyytop: To avoid ambiguity I updated the error message as blow:
```
The BTC UTXO must not be bound to multiple CKB cells
```

packages/ckb/src/spore/leap.ts
Flouse: After discussion, https://github.com/ckb-cell/rgbpp-sdk/pull/130/commits/529e68c1a1593c1e711d90d30f2d617851105509 renamed `isDataEmpty ` to `isDataMustBeEmpty`, which is more clear.

Flouse (approved):

Dawn-githup:
https://github.com/ckb-cell/rgbpp-sdk/blob/35cd592ea57c6e1f172be521282c806a20dedeaf/examples/rgbpp/spore/6-unlock-btc-time-cell.ts#L38

Dawn-githup:
develop branch regression verification:
https://inquisitive-puma-8a7.notion.site/Support-Leap-spores-and-add-examples-test-5c5cfbff25674154a2315ef93eab9d0e

duanyytop:
Fixed in https://github.com/ckb-cell/rgbpp-sdk/pull/145

PR #133, opened 2024-04-24
feat(btc): simplify inner logic of the sendBtc API

ShookLyngs:
## Changes
- Simplify the inner logic of the sendBtc() API with the createSendUtxosBuilder() API in the @rgbpp-sdk/btc

### DIFF
--- a/packages/btc/src/api/sendBtc.ts
+++ b/packages/btc/src/api/sendBtc.ts
@@ -1,6 +1,7 @@
 import { bitcoin } from '../bitcoin';
 import { DataSource } from '../query/source';
 import { InitOutput, TxBuilder } from '../transaction/build';
+import { createSendUtxosBuilder } from './sendUtxos';
 
 export interface SendBtcProps {
   from: string;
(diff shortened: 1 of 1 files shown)


packages/btc/src/api/sendBtc.ts
```diff
@@ -18,31 +19,23 @@ export async function createSendBtcBuilder(props: SendBtcProps): Promise<{
   feeRate: number;
   fee: number;
 }> {
-  const tx = new TxBuilder({
+  // By default, all outputs in the sendBtc() API are fixed
+  const outputs = props.tos.map((to) => ({
+    fixed: true,
+    ...to,
```
duanyytop, line 25: If the `to` contains a `fixed` parameter, and the value is false, then fixed will eventually be false

ShookLyngs: Yes, this is intentional, allowing users to alter the default value of "fixed" if needed.

duanyytop (approved):

Flouse (approved):

PR #136, opened 2024-04-25
fix(btc): error message when reporting INSUFFICIENT_UTXO

ShookLyngs:
## Changes
1. Add an `allowInsufficient` option (set to `true` by default) to the DataSource.collectSatoshi() API to allow the collection process to be completed when the balance of the target address is insufficient, instead of throwing the INSUFFICIENT_UTXO error
2. Fix the expected amount of error when collected satoshi and the balance of the target address are both insufficient for a change output to be generated in the transaction (reported by @Dawn-githup)

### DIFF
--- a/packages/btc/src/transaction/build.ts
+++ b/packages/btc/src/transaction/build.ts
@@ -305,7 +305,11 @@ export class TxBuilder {
     // If not collected enough satoshi, throw error
     const insufficientBalance = collected < targetAmount;
     if (insufficientBalance) {
-      throw TxBuildError.withComment(ErrorCodes.INSUFFICIENT_UTXO, `expected: ${targetAmount}, actual: ${collected}`);
+      const recommendedDeposit = collected - targetAmount + this.minUtxoSatoshi;
+      throw TxBuildError.withComment(
+        ErrorCodes.INSUFFICIENT_UTXO,
+        `expected: ${targetAmount}, actual: ${collected}. You may wanna deposit more satoshi to prevent the error, for example: ${recommendedDeposit}`,
+      );
     }
     const insufficientForChange = changeAmount > 0 && changeAmount < this.minUtxoSatoshi;
     if (insufficientForChange) {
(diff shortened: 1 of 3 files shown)


Flouse (commented):
Please update `packages/btc/README.md`.

https://github.com/ckb-cell/rgbpp-sdk/blob/53ecb2e950ac11b2c30e1dfc2f409fce53917ff9/packages/btc/README.md#L368-L378

Flouse (approved):

Dawn-githup:
https://inquisitive-puma-8a7.notion.site/3cebd9c87be54298805f6bfa5106342e

packages/btc/src/transaction/build.ts
```diff
@@ -307,7 +308,7 @@ export class TxBuilder {
     }
     const insufficientForChange = changeAmount > 0 && changeAmount < this.minUtxoSatoshi;
     if (insufficientForChange) {
-      const shiftedExpectAmount = targetAmount + changeUtxoNeedAmount;
+      const shiftedExpectAmount = collected + changeUtxoNeedAmount;
```
Dawn-githup, line 311: The expected result here will be more than the "insufficientBalance" expected result;
```
if (insufficientBalance) {
      throw TxBuildError.withComment(ErrorCodes.INSUFFICIENT_UTXO, `expected: ${targetAmount}, actual: ${collected}`);
    }
```

Here might be some confusion: initially, the funds are insufficient, prompting users to top up. However, upon returning after the top-up, they still find the balance insufficient.

Dawn-githup:
https://inquisitive-puma-8a7.notion.site/error-message-when-reporting-INSUFFICIENT_UTXO-verify-3cebd9c87be54298805f6bfa5106342e

packages/btc/src/transaction/build.ts
ShookLyngs: Updated the error message when insufficient balance. Now it should warn the user to deposit more satoshi in order to prevent the change UTXO value validation.

Dawn-githup (approved):

PR #138, opened 2024-04-25
refactor(service|btc): use org mempool api

ShookLyngs:
## Changes
1. Use `cell.mempool.space` as the Mempool API hostname
2. Allow some btc/service tests to retry more times: 3 -> 5 times
3. Updated some tests to prevent using the Mempool API too much

### DIFF
--- a/packages/btc/src/query/mempool.ts
+++ b/packages/btc/src/query/mempool.ts
@@ -10,11 +10,11 @@ export interface MempoolConfig {
 
 export const mempoolConfigs: Record<'mainnet' | 'testnet', MempoolConfig> = {
   testnet: {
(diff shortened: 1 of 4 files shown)


ShookLyngs:
The 503 error still exists after the changes, it may require more effort to resolve the issue.
I also updated some tests to avoid accessing the Mempool API, or just give it more tries when failed.

ahonn:
Maybe you can consider using https://vitest.dev/guide/mocking.html#modules to mock `@mempool/mempool.js`, and then wait for the Assets API to be connected to mempool.space before replacing it.

ShookLyngs:
Closing the PR, the PR can be replaced by #148 because it's a more reliable update.

PR #149, opened 2024-04-29
feat(service): throw service error with context

### LINKED ISSUE #95: Inject context into errors of rgbpp-sdk/service
- Refer to: https://github.com/ckb-cell/rgbpp-sdk/pull/91#discussion_r1546553599

---
- similar issue: https://github.com/ckb-cell/rgbpp-sdk/issues/95

ShookLyngs:
## Changes
- Add `context` in the error object of the service lib
- Add error handling recipe in service lib readme

## Related issues
- Resolves #95

## Details

You can now find context in the `BtcAssetsApiError` if the error was thrown from a request:
```ts
try { 
  ... 
} catch (e) {
  if (e instanceof BtcAssetsApiError) {
    console.log(e.context?.request);
    console.log(e.context?.response);
  }
}
```

This is the type of `e.context`:
```ts
interface BtcAssetsApiContext {
  request: {
    url: string;
    body?: Record<string, any>;
    params?: Record<string, any>;
  };
  response: {
    status: number;
    data?: Record<string, any> | string;
  };
}
```

### DIFF
--- a/packages/service/README.md
+++ b/packages/service/README.md
@@ -85,6 +85,28 @@ console.log(res);
 
 All available APIs in the [BtcAssetsApi](#btcassetsapi-1) section.
 
+### Handling service errors
+
+You can identify the error by its `code` and `message`, or by its detailed `context`:
+
+```ts
+import { BtcAssetsApiError, ErrorCodes } from '@rgbpp-sdk/service';
+
+try {
+...
+} catch (e) {
+  if (e instanceof BtcAssetsApiError) {
+    // error code
+    console.log(e.code === ErrorCodes.ASSETS_API_UNAUTHORIZED);
+    // error message
+    console.log(e.message);
+    // detailed context info  (if thrown from a request)
+    console.log(e.context?.request);
+    console.log(e.context?.response);
+  }
+}
+```
+
 ## Types
 
 ### BtcAssetsApi
--- a/packages/service/README.md
+++ b/packages/service/README.md
@@ -96,13 +96,25 @@ try {
 ...
 } catch (e) {
   if (e instanceof BtcAssetsApiError) {
-    // error code
-    console.log(e.code === ErrorCodes.ASSETS_API_UNAUTHORIZED);
-    // error message
-    console.log(e.message);
-    // detailed context info  (if thrown from a request)
-    console.log(e.context?.request);
-    console.log(e.context?.response);
+    // check error code
+    console.log(e.code === ErrorCodes.ASSETS_API_UNAUTHORIZED); // true
+    // print the whole error
+    console.log(JSON.stringify(e));
+    /*{
+      "message": "BtcAssetsAPI unauthorized, please check your token/origin: (401) Authorization token is invalid: The token header is not a valid base64url serialized JSON.",
+      "code": 2,
+      "context": {
+        "request": {
+          "url": "https://btc-assets-api.url/bitcoin/v1/info"
+        },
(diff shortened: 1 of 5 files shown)


packages/service/src/service/base.ts
```diff
@@ -35,7 +35,8 @@ export class BtcAssetsApiBase implements BaseApis {
     const packedParams = params ? '?' + new URLSearchParams(pickBy(params, (val) => val !== undefined)).toString() : '';
```
duanyytop, line 35: `val => !! val` is more simple and clear?

packages/service/README.md
```diff
@@ -85,6 +85,28 @@ console.log(res);
...
+try {
+...
+} catch (e) {
+  if (e instanceof BtcAssetsApiError) {
+    // error code
+    console.log(e.code === ErrorCodes.ASSETS_API_UNAUTHORIZED);
+    // error message
+    console.log(e.message);
+    // detailed context info  (if thrown from a request)
+    console.log(e.context?.request);
```
duanyytop, line 104: I think it is possible to add `JSON.string()` to get the expected log information, and maybe you can have a try locally.

duanyytop (approved):

packages/service/src/service/base.ts
ahonn: I think `val !== undefined` is better, if val is 0, it will be filtered out here

duanyytop: Yes. The type of val is unknown

packages/service/src/service/base.ts
```diff
@@ -126,3 +138,11 @@ export class BtcAssetsApiBase implements BaseApis {
     this.token = token.token;
   }
 }
+
+function tryParseBody(body: unknown): Record<string, any> | undefined {
+  try {
+    return typeof body === 'string' ? JSON.parse(body) : void 0;
+  } catch {
+    return void 0;
```
ahonn, line 146: We should avoid using `void 0`, it is the same as undefined, but not as easy to understand.

Some libraries/packages use `void 0` because `undefined` can be shadowed by local variables, and "void 0" is shorter and can save a few bytes. 
But for modern JavaScript/TypeScript, this is a problem that does not require concern. Local variable shadowing cannot occur in TypeScript, and bundle size should be left to webpack/esbuild or other tools, rather than written as `void 0`

duanyytop: I agree that undefined is better than void 0

ShookLyngs: It's kinda a coding style habit for me, I use it because it's easy to type. 
I'll replace all `void 0` to `undefined` in another pull request.

packages/service/README.md
ShookLyngs: Added a JSON print in the readme, now it should be clearer.

ahonn (approved):

packages/service/src/service/base.ts
```diff
@@ -56,8 +57,19 @@ export class BtcAssetsApiBase implements BaseApis {
       // do nothing
     }
 
-    const status = res.status;
     let comment: string | undefined;
+    const status = res.status;
+    const context: BtcAssetsApiContext = {
+      request: {
+        url: `${this.url}${route}${packedParams}`,
```
Flouse, line 64:
```suggestion
        url,
```

Flouse (approved):

PR #151, opened 2024-05-06
refactor: use ESLint globally in the repo

### LINKED ISSUE #131: [code-style] Add @typescript-eslint/recommended configuration to the project
## Description
The rgbpp-sdk project currently does not have a comprehensive ESLint configuration set up. To improve code quality and maintainability, I propose adding the @typescript-eslint/recommended configuration to the project.

@ShookLyngs @duanyytop @yuche Do you have a better suggestion?

## Benefits
- Enforces a consistent code style and best practices for this project
- Catches common TypeScript-specific errors and potential issues early in the development process
- Improves code maintainability and readability

ShookLyngs:
## Changes
- Add `ESLint` in the root of the repo
- Reformat the code of the service/btc/ckb lib with the `ESLint`
- Refactor the examples to use global `ESLint` config and format rules
- Use `lint` command in packages/examples to check code in the `test.yaml` (instead of fixing)
- Tune the timeout of btc/service lib to allow longer tests (mempool-based APIs take longer to process)

## Related issues
- Resolves #131

### DIFF
--- a/examples/rgbpp/package.json
+++ b/examples/rgbpp/package.json
@@ -6,7 +6,8 @@
   "type": "module",
   "scripts": {
     "format": "prettier --write '**/*.{js,ts}'",
-    "lint:fix": "pnpm eslint . && prettier --check '**/*.{js,ts}'"
+    "lint": "tsc && eslint . && prettier --check '**/*.{js,ts}'",
+    "lint:fix": "tsc && eslint --fix --ext .js,.ts . && prettier --write '**/*.{js,ts}'"
   },
   "dependencies": {
     "@exact-realty/multipart-parser": "^1.0.13",
@@ -18,10 +19,7 @@
     "axios": "^1.6.8"
   },
   "devDependencies": {
-    "@eslint/js": "^9.1.1",
     "@types/node": "^20.11.28",
-    "eslint": "^8.56.0",
-    "typescript": "^5.4.2",
-    "typescript-eslint": "^7.7.1"
+    "typescript": "^5.4.2"
   }
 }
--- a/package.json
+++ b/package.json
@@ -7,10 +7,10 @@
   "scripts": {
     "prepare": "husky",
     "test:packages": "turbo run test --filter=./packages/*",
-    "build": "turbo run build",
     "build:packages": "turbo run build --filter=./packages/*",
     "lint:fix": "turbo run lint:fix",
-    "lint:fix-all": "prettier --write '{packages,apps, examples}/**/*.{js,jsx,ts,tsx,md,json}'",
+    "lint:packages": "turbo run lint --filter=./{packages,examples}/*",
+    "lint:fix-all": "prettier --write '{packages,apps,examples}/**/*.{js,jsx,ts,tsx,md,json}'",
     "clean": "turbo run clean",
     "clean:packages": "turbo run clean --filter=./packags/*",
     "clean:dependencies": "pnpm clean:sub-dependencies && rimraf node_modules",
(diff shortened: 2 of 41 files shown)


examples/rgbpp/package.json
```diff
@@ -6,7 +6,8 @@
   "type": "module",
   "scripts": {
     "format": "prettier --write '**/*.{js,ts}'",
-    "lint:fix": "pnpm eslint . && prettier --check '**/*.{js,ts}'"
+    "lint": "tsc && eslint . && prettier --check '**/*.{js,ts}'",
+    "lint:fix": "tsc && eslint --fix --ext .js,.ts . && prettier --write '**/*.{js,ts}'"
```
duanyytop, line 10: I don't think the `tsc` of the examples is needed

package.json
```diff
@@ -7,10 +7,10 @@
   "scripts": {
     "prepare": "husky",
     "test:packages": "turbo run test --filter=./packages/*",
-    "build": "turbo run build",
     "build:packages": "turbo run build --filter=./packages/*",
     "lint:fix": "turbo run lint:fix",
-    "lint:fix-all": "prettier --write '{packages,apps, examples}/**/*.{js,jsx,ts,tsx,md,json}'",
+    "lint:packages": "turbo run lint --filter=./{packages,examples}/*",
+    "lint:fix-all": "prettier --write '{packages,apps,examples}/**/*.{js,jsx,ts,tsx,md,json}'",
```
duanyytop, line 13: I suggest to rename the `lint:fix-all` to `format`, because the `prettier` is different from `eslint` and `typescript/eslint`

examples/rgbpp/package.json
duanyytop: After discussing with shook, I think tsc is needed @ShookLyngs

duanyytop (approved):

Flouse (approved):

packages/btc/src/utils.ts
```diff
@@ -11,14 +15,14 @@ function tapTweakHash(publicKey: Buffer, hash: Buffer | undefined): Buffer {
...
   signer: T,
   options?: {
     network?: bitcoin.Network;
     tweakHash?: Buffer;
   },
 ): bitcoin.Signer {
-  let privateKey: Uint8Array | undefined = (signer as any).privateKey;
+  let privateKey: Uint8Array | undefined = signer.privateKey ? new Uint8Array(signer.privateKey) : undefined;
   if (!privateKey) {
     throw new Error('Private key is required for tweaking signer!');
```
ahonn, line 27: By exchanging these two lines, we don't need to check whether it is undefined in `let privateKey`, and the type can also be determined to be Uint8Array.

packages/ckb/src/utils/case-parser.ts
```diff
@@ -1,10 +1,10 @@
...
-export const toCamelcase = (object: any) => {
+export const toCamelcase = (object: unknown) => {
   try {
     return JSON.parse(
       JSON.stringify(
-        camelcaseKeys(object, {
+        camelcaseKeys(object!, {
           deep: true,
         }),
       ),
```
ahonn, line 10: I don’t think it’s a good idea to do `JSON.parse` after `JSON.stringify`. It seems like this is just to convert object keys to camelcase. Is it possible to just use `camelcaseKeys`? I don’t quite understand the reason why use stringify and parse here.

duanyytop: There is indeed a performance loss, and a more reasonable approach is to use generics to implement type conversion.

I updated the `toCamelcase` function and related code in https://github.com/ckb-cell/rgbpp-sdk/pull/151/commits/aed41d85101aa2dd4031c561deaef9e24ef56296

cc @ahonn

ahonn (approved):

PR #164, opened 2024-05-10
chore(btc/service/apps/ci): rename env vars "SERVICE_*" -> "BTC_SERVICE_*"

ShookLyngs:
## Changes
- Rename env vars `SERVICE_*` -> `BTC_SERVICE_*` (requested by @duanyytop)

PR #165, opened 2024-05-11
style(btc/service): replace all "void 0" to "undefined"

ShookLyngs:
## Changes
- Replace all "void 0" to "undefined" in the btc/service lib (code style update suggested by @ahonn at https://github.com/ckb-cell/rgbpp-sdk/pull/149#discussion_r1584155981)

## Checklist
- [x] Merge #164

### DIFF
--- a/packages/btc/src/address.ts
+++ b/packages/btc/src/address.ts
@@ -28,7 +28,7 @@ export function isSupportedFromAddress(address: string) {
  */
 export function publicKeyToPayment(publicKey: string, addressType: AddressType, networkType: NetworkType) {
   if (!publicKey) {
-    return void 0;
+    return undefined;
   }
 
   const network = networkTypeToNetwork(networkType);
@@ -64,7 +64,7 @@ export function publicKeyToPayment(publicKey: string, addressType: AddressType, 
     });
   }
 
-  return void 0;
+  return undefined;
 }
 
 /**
@@ -147,7 +147,7 @@ export function decodeAddress(address: string): {
           addressType = AddressType.P2TR;
         }
       }
-      if (networkType !== void 0 && addressType !== void 0) {
+      if (networkType !== undefined && addressType !== undefined) {
         return {
           networkType,
           addressType,
@@ -182,7 +182,7 @@ export function decodeAddress(address: string): {
         addressType = AddressType.P2SH_P2WPKH;
       }
 
-      if (networkType !== void 0 && addressType !== void 0) {
+      if (networkType !== undefined && addressType !== undefined) {
         return {
           networkType,
           addressType,
--- a/packages/btc/src/query/source.ts
+++ b/packages/btc/src/query/source.ts
@@ -35,14 +35,14 @@ export class DataSource {
     const txId = remove0x(hash);
     const tx = await this.service.getBtcTransaction(txId);
     if (!tx) {
-      return void 0;
+      return undefined;
     }
     if (requireConfirmed && !tx.status.confirmed) {
       throw TxBuildError.withComment(ErrorCodes.UNCONFIRMED_UTXO, `hash: ${hash}, index: ${index}`);
     }
     const vout = tx.vout[index];
     if (!vout) {
-      return void 0;
+      return undefined;
     }
 
     const scriptBuffer = Buffer.from(vout.scriptpubkey, 'hex');
--- a/packages/service/src/service/base.ts
+++ b/packages/service/src/service/base.ts
@@ -33,8 +33,8 @@ export class BtcAssetsApiBase implements BaseApis {
     }
 
     const packedParams = params ? '?' + new URLSearchParams(pickBy(params, (val) => val !== undefined)).toString() : '';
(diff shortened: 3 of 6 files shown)


packages/btc/src/address.ts
```diff
@@ -28,7 +28,7 @@ export function isSupportedFromAddress(address: string) {
  */
 export function publicKeyToPayment(publicKey: string, addressType: AddressType, networkType: NetworkType) {
   if (!publicKey) {
-    return void 0;
+    return undefined;
```
duanyytop, line 31: It should be better to throw an exception than return `undefined`, so that the caller can realize the problem early.

packages/btc/src/address.ts
```diff
@@ -64,7 +64,7 @@ export function publicKeyToPayment(publicKey: string, addressType: AddressType,
     });
   }
 
-  return void 0;
+  return undefined;
```
duanyytop, line 67: Same as above

packages/service/src/service/base.ts
```diff
@@ -97,7 +97,7 @@ export class BtcAssetsApiBase implements BaseApis {
       throw BtcAssetsApiError.withComment(ErrorCodes.ASSETS_API_RESPONSE_ERROR, comment, context);
     }
     if (status !== 200) {
-      return void 0 as T;
+      return undefined as T;
```
duanyytop, line 100: Same as above

packages/service/src/service/base.ts
```diff
@@ -33,8 +33,8 @@ export class BtcAssetsApiBase implements BaseApis {
     }
 
     const packedParams = params ? '?' + new URLSearchParams(pickBy(params, (val) => val !== undefined)).toString() : '';
-    const withOriginHeaders = this.origin ? { origin: this.origin } : void 0;
-    const withAuthHeaders = requireToken && this.token ? { Authorization: `Bearer ${this.token}` } : void 0;
+    const withOriginHeaders = this.origin ? { origin: this.origin } : undefined;
```
duanyytop, line 36:
```suggestion
    const withOriginHeaders = { origin: this.origin };
```

packages/btc/src/query/source.ts
```diff
@@ -35,14 +35,14 @@ export class DataSource {
     const txId = remove0x(hash);
     const tx = await this.service.getBtcTransaction(txId);
     if (!tx) {
-      return void 0;
+      return undefined;
```
duanyytop, line 38: It is recommended to throw an exception directly here

packages/btc/src/query/source.ts
```diff
@@ -35,14 +35,14 @@ export class DataSource {
...
     }
     if (requireConfirmed && !tx.status.confirmed) {
       throw TxBuildError.withComment(ErrorCodes.UNCONFIRMED_UTXO, `hash: ${hash}, index: ${index}`);
     }
     const vout = tx.vout[index];
     if (!vout) {
-      return void 0;
+      return undefined;
```
duanyytop, line 45: It is recommended to throw an exception directly here

packages/btc/src/address.ts
ShookLyngs: Yes, this part of the code is not aligned with the the other functions in the `address.ts`. Some return `undefined` while some don't, I'll check them out and modify in another PR, if nessesary.

packages/service/src/service/base.ts
ShookLyngs: This part is expected but the if statement was not clear, I'll update it in another PR.

packages/btc/src/query/source.ts
ShookLyngs: I looked at the reference of the function and it does look weird to throw afterward, I'll re-think about the signature of the function.

packages/service/src/service/base.ts
duanyytop: If the origin is `undefined`, the origin field will not appear in the header and you can have a try.

ShookLyngs: I split them for better readability, although it may have failed to achieve its job. Another solution is that these properties can also be packed into the fetch call:
```ts
fetch('xxx', {
  headers: {
    authorization: requireToken && this.token ? `Bearer ${this.token}` : undefined,
    origin: this.origin,
    ...headers,
  },
})
```

duanyytop: LGTM

ahonn:
Should be rebase from the developer branch to opt-out of the code changes by https://github.com/ckb-cell/rgbpp-sdk/pull/164.

ShookLyngs:
> This PR should be rebase from the develop branch to opt-out of the #164 code changes.

The commit message was altered when the #164 was merged. It's now correctly rebased.

duanyytop:
After discussion,  throwing errors instead of returning `undefined` will be updated in another PR.

ahonn (approved):

Flouse (approved):

PR #166, opened 2024-05-11
fix(btc): message of INSUFFICIENT_UTXO

ShookLyngs:
## Changes
- Fix the message of INSUFFICIENT_UTXO when `collected < targetAmount` (@Dawn-githup)


Dawn-githup (approved):

Flouse (approved):
### Previous Error Msg
Insufficient UTXO to construct the transaction: expected: 12546, actual: 0. You may wanna deposit more satoshi to prevent the error, for example: `-11346`

---

Should be fixed by this PR.

PR #168, opened 2024-05-11
fix(ckb/btc): Check rgbpp capacity for change and paymaster cell

duanyytop:
### CKB package
- Add `isRgbppCapacitySufficientForChange` function to check whether the capacity of inputs is sufficient for outputs
- Return the rest of the capacity to sender when the RGB++ change cell is needed
- Determine whether a paymaster cell is needed based on whether the capacity of the inputs is sufficient for the outputs
- Add `checkCkbTxInputsCapacitySufficient` function for BTC package to determine whether a paymaster cell is needed

### BTC package
- Update paymaster cell logic with new function `checkCkbTxInputsCapacitySufficient`

### DIFF
--- a/packages/ckb/src/utils/ckb-tx.ts
+++ b/packages/ckb/src/utils/ckb-tx.ts
@@ -11,6 +11,8 @@ import {
 import { Hex, RgbppTokenInfo } from '../types';
 import { PERSONAL, blake2b, hexToBytes, serializeInput, serializeScript } from '@nervosnetwork/ckb-sdk-utils';
 import { encodeRgbppTokenInfo, genBtcTimeLockScript } from './rgbpp';
+import { Collector } from '../collector';
+import { NoLiveCellError } from '../error';
 
 export const calculateTransactionFee = (txSize: number, feeRate?: bigint): bigint => {
   const rate = feeRate ?? BigInt(1100);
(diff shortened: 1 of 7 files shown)


packages/ckb/src/utils/rgbpp.ts
```diff
@@ -253,3 +254,11 @@ export const throwErrorWhenRgbppCellsInvalid = (
     throw new NoRgbppLiveCellError('No rgbpp cells found with the xudt type script and the rgbpp lock args');
   }
 };
+
+export const isRgbppCapacitySufficientForChange = (
```
Flouse, line 258:
```suggestion
/**
 * Check if the tx's unoccupied capacity is enough to create a new rgbpp-cell as a UDT change cell
 * 
 * @param sumUdtInputsCapacity 
 * @param receiverOutputCapacity 
 * @returns 
 */
export const isRgbppCapacitySufficientForChange = (```

**Review by Flouse — APPROVED** [review]

**Review by ahonn — APPROVED** [review]

File: packages/ckb/src/utils/ckb-tx.ts [hunk 2]
```diff
@@ -139,3 +141,27 @@ export const isScriptEqual = (s1: Hex | CKBComponents.Script, s2: Hex | CKBCompo
...
+ */
+export const checkCkbTxInputsCapacitySufficient = async (
+  ckbTx: CKBComponents.RawTransaction,
+  collector: Collector,
+): Promise<boolean> => {
+  let sumInputsCapacity = BigInt(0);
+  for await (const input of ckbTx.inputs) {
+    const liveCell = await collector.getLiveCell(input.previousOutput!);
+    if (!liveCell) {
+      throw new NoLiveCellError('The cell with the specific out point is dead');
```
ShookLyngs, line 159: Is it the same if returning `false` when the cell is not lived? 
Didn't expect that the user has to wrap a try catch scope by looking at the signature.

ShookLyngs: After discussion, we think throwing an error here can help users identify internal errors. 
The btc lib doesn't catch this function, and just leaves it to the users.

ShookLyngs (approved):

PR #171, opened 2024-05-15
feat(rgbpp-sdk/ckb): Build ckb raw tx for spore creation

duanyytop:
## Main Changes

- Add `buildAppendingIssuerCellToSporesCreateTx` function to build ckb raw tx to be signed for spores creation
- The Dapp can use this function to build ckb raw tx and let the JoyID wallet or other wallets to sign the tx

### DIFF
--- a/packages/ckb/src/spore/spore.ts
+++ b/packages/ckb/src/spore/spore.ts
@@ -9,7 +9,7 @@ import {
 import { buildPreLockArgs, calculateCommitment, genRgbppLockScript } from '../utils/rgbpp';
 import {
   AppendIssuerCellToSporeCreate,
-  BuildAppendingIssuerCellToSporeCreateTx,
+  BuildAppendingIssuerCellTxParams,
   CreateSporeCkbVirtualTxParams,
   Hex,
   SporeCreateVirtualTxResult,
@@ -170,7 +170,7 @@ export const buildAppendingIssuerCellToSporesCreateTx = async ({
   sumInputsCapacity,
   witnessLockPlaceholderSize = SECP256K1_WITNESS_LOCK_SIZE,
   ckbFeeRate,
-}: BuildAppendingIssuerCellToSporeCreateTx): Promise<CKBComponents.RawTransactionToSign> => {
+}: BuildAppendingIssuerCellTxParams): Promise<CKBComponents.RawTransactionToSign> => {
   const isMainnet = issuerAddress.startsWith('ckb');
   const rawTx = ckbRawTx as CKBComponents.RawTransactionToSign;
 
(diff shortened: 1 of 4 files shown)


packages/ckb/src/types/spore.ts
```diff
@@ -51,6 +51,21 @@ export interface SporeCreateVirtualTxResult {
   clusterCell: IndexerCell;
 }
 
+export interface BuildAppendingIssuerCellToSporeCreateTx {
```
ShookLyngs, line 54: I see the other types here are either suffixed with `Params` or `Result`, only the issuer cell related signatures are exceptions. Are they by design? The exception types: `BuildAppendingIssuerCellToSporeCreateTx`, `AppendIssuerCellToSporeCreate`

duanyytop: This is a historical issue. In order to maintain compatibility, it is recommended not to change it.

duanyytop: I rename the `BuildAppendingIssuerCellToSporeCreateTx` to `BuildAppendingIssuerCellTxParams` in https://github.com/ckb-cell/rgbpp-sdk/pull/171/commits/18a1c387c6a774028f7e9dab15fcdad0e7cc5bbf.

I suggest keeping the `AppendIssuerCellToSporeCreate` not changed

ShookLyngs (approved):

ahonn (approved):

packages/ckb/src/spore/spore.ts
```diff
@@ -205,13 +202,47 @@ export const appendIssuerCellToSporesCreate = async ({
   rawTx.outputs = [...rawTx.outputs, changeOutput];
   rawTx.outputsData = [...rawTx.outputsData, '0x'];
 
-  rawTx.cellDeps = [...rawTx.cellDeps, getSecp256k1CellDep(isMainnet)];
+  if (witnessLockPlaceholderSize === SECP256K1_WITNESS_LOCK_SIZE) {
+    rawTx.cellDeps = [...rawTx.cellDeps, getSecp256k1CellDep(isMainnet)];
```
Flouse, line 206: Perhaps this line of code could be moved to `fn appendIssuerCellToSporesCreate`:L242

Flouse (approved):

packages/ckb/src/spore/spore.ts
duanyytop: > Perhaps this line of code could be moved to fn appendIssuerCellToSporesCreate:L242

This will lead to inaccurate fee estimates.

Judging which lock it belongs to based on the issuer lock seems to be a better choice

duanyytop: Move the secp256k1 cell dep to signing stage in https://github.com/ckb-cell/rgbpp-sdk/pull/171/commits/ce21ad377477b13e3d25c4d52767ebf1c32ed2b1

PR #176, opened 2024-05-16
refactor(btc): add notes to TxBuilder.payFee() related functions

### LINKED ISSUE #140: Detailed explanation on TxBuilder.payFee()
The TxBuilder.payFee() method is the core part in BTC transaction construction. However, the method is currently not self-explanatory enough for the readers, in other words, it's hard to read and understand. So we'd better add more detailed comments to the method.

ShookLyngs:
## Changes
- Add notes/comments to better explain the inner logic of the `TxBuilder.payFee()` related functions
    > The change doesn't include actual logic refactors, therefore no changeset was added

## Related issues
- Resolves #140

### DIFF
--- a/packages/btc/src/transaction/build.ts
+++ b/packages/btc/src/transaction/build.ts
@@ -145,17 +145,15 @@ export class TxBuilder {
     const originalInputs = clone(this.inputs);
     const originalOutputs = clone(this.outputs);
 
-    // Fetch the latest average fee rate if feeRate param is not provided
-    // The transaction is expected be confirmed within half an hour with the fee rate
-    let averageFeeRate: number | undefined;
+    // Fill a default recommended fee rate if props.feeRate is not provided
+    let defaultFeeRate: number | undefined;
     if (!feeRate && !this.feeRate) {
       const feeRates = await this.source.service.getBtcRecommendedFeeRates();
-      averageFeeRate = feeRates.fastestFee;
+      defaultFeeRate = feeRates.fastestFee;
     }
 
-    // Use the feeRate param if it is specified,
-    // otherwise use the average fee rate from the DataSource
-    const currentFeeRate = feeRate ?? this.feeRate ?? averageFeeRate!;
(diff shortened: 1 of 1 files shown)


packages/btc/src/transaction/build.ts
```diff
@@ -362,6 +375,13 @@ export class TxBuilder {
...
 
     if (amount < this.minUtxoSatoshi) {
+      // If the change is not enough to create a non-dust output, try collect more.
+      // - injectCollected=true, expect to put all (collected + amount) of satoshi as change
+      // - deductFromOutputs=false, do not collect satoshi from the outputs
+      // An example:
+      // 1. Expected to return change of 500 satoshi, amount=500
+      // 2. Collected 2000 satoshi from the "fromAddress", collected=2000
+      // 3. Create a change output and return (collected + amount), output.value=2000+500=2500
       const { collected } = await this.injectSatoshi({
```
Flouse, line 385:
```suggestion
      // If the change is not enough to create a non-dust output, try collect more.
      // - injectCollected=true, expect to put all (collected + amount) of satoshi as change
      // - deductFromOutputs=false, do not collect satoshi from the outputs
      // An example:
      // 1. Expected to return change of 500 satoshi, amount=500
      // 2. Collected 2000 satoshi from the "fromAddress", collected=2000
      // 3. Create a change output and return (collected + amount), output.value=2000+500=2500
      //
      // TODO(Low-Priority):
      // If the cost of creating a non-dust output is greater than the change amount, it may be
      // reasonable to just pay the change amount as an additional fee to the miners.
      const { collected } = await this.injectSatoshi({
```

Flouse (approved):

packages/btc/src/transaction/build.ts
ShookLyngs: Raised an issue for the suggestion instead: https://github.com/ckb-cell/rgbpp-sdk/issues/183

PR #181, opened 2024-05-20
feat: add no_cache params to btc/rgbpp service api

ahonn:
## Changes
The assets service will cache Bitcoin's UTXO data and RGB++'s cell data in subsequent updates and provide a `no_cache` parameter to skip cache. This PR adds parameter types and corresponding unit tests to support it.

Issue #173, opened 2024-05-15
Building BTC TX with an address having excessive UTXOs takes forever


ShookLyngs:
## Issue

A transaction with specific `from` address can take forever to build, for example:
```ts
await sendBtc({
  from: 'tb1quqtqsh5jrlr9p5wnpu3rs883lqh4avpwc766x3',
  tos: [
    {
      address: 'tb1quqtqsh5jrlr9p5wnpu3rs883lqh4avpwc766x3',
      value: 1000,
    },
  ],
  source,
});
```

What happens:
1. The construction process contains a loop to estimate the fee of the transaction
2. In each loop, the `BtcAssetsApi.getBtcUtxos()` API is called to get the latest UTXO[] list
3. Each request can take 8-15 seconds, and in each loop, more UTXOs are collected to the inputs
4. If the inputs length increased, the estimated fee is also increased, causing another round of the loop
5. The condition to break the loop takes a very long time to be satisfied (due to the INSUFFICIENT_UTXO error)
6. The loop (tested by @Dawn-githup) can take 40+ minutes to fail, which is clearly abnormal

## Investigation & resolvers
- [x] Should the `BtcAssetsApi` to be queried once per build to avoid long waiting time of requests?
    - The UTXO[] list can be updated during the construction, should we worry about it?
- [ ] There's only a 1000 satoshi output in the outputs, not a huge cost at all, so why won't the loop stop?
    - Is it because some bugs remain in the pay fee process?

ahonn:
> Should the BtcAssetsApi to be queried once per build to avoid long waiting time of requests?

Yes, we should query once per build and check the inputs UTXO after the transaction is built.
The `BtcAssetsApi` can provide a new API `/bitcoin/v1/tx/:txid/outspend/:vout` to return the spending status of each UTXO. It would be faster and more efficient.

ShookLyngs:
## Debug report

Ran some tests today:

1. For the example mentioned above, the total loop should be 6 times.
2. In each loop, the `BtcAssetsApi.getBtcUtxos()` API is used for querying UTXO[].
3. In each loop, the `BtcAssetsApi.getRgbppAssetsByBtcUtxo()` is used for checking each input's corresponding RGBPP assets, and if the UTXO is bound with any, it's excluded from the collection process.

[...]

PR #184, opened 2024-05-23
feat(btc|service): support internal query data caching in TxBuilder.payFee()

ShookLyngs:
## Changes
- Support query data caching internally in TxBuilder/DataSource, preventing query from the service too often
- Add type changes in the #181 to the service readme

## Query data cache

Previously, when collecting inputs for paying fees, the `BtcAssetsApi.getBtcUtxos()` API and the `BtcAssetsApi.getRgbppAssetsByBtcUtxo()` API were queried multiple times. In this PR, a DataCache is added to prevent querying the same API with the same params too often.

Reference: https://github.com/ckb-cell/rgbpp-sdk/issues/173#issuecomment-2112523137

## Test
- [x] @Dawn-githup

### DIFF
--- a/packages/btc/src/query/source.ts
+++ b/packages/btc/src/query/source.ts
@@ -105,8 +105,8 @@ export class DataSource {
     allowInsufficient?: boolean;
     onlyNonRgbppUtxos?: boolean;
     onlyConfirmedUtxos?: boolean;
-    noUtxosCache?: boolean;
-    cacheKey?: string;
+    noAssetsApiCache?: boolean;
+    internalCacheKey?: string;
     excludeUtxos?: {
       txid: string;
       vout: number;
@@ -122,19 +122,19 @@ export class DataSource {
       minUtxoSatoshi,
       onlyConfirmedUtxos,
       onlyNonRgbppUtxos,
-      noUtxosCache,
-      cacheKey,
+      noAssetsApiCache,
+      internalCacheKey,
       allowInsufficient = false,
       excludeUtxos = [],
     } = props;
 
     const utxos = await this.cache.optionalCacheUtxos({
-      key: cacheKey,
+      key: internalCacheKey,
       getter: () =>
         this.getUtxos(address, {
           only_confirmed: onlyConfirmedUtxos,
           min_satoshi: minUtxoSatoshi,
-          no_cache: noUtxosCache,
+          no_cache: noAssetsApiCache,
         }),
     });
 
--- a/packages/btc/src/query/source.ts
+++ b/packages/btc/src/query/source.ts
@@ -6,14 +6,17 @@ import { TxAddressOutput } from '../transaction/build';
 import { isOpReturnScriptPubkey } from '../transaction/embed';
 import { addressToScriptPublicKeyHex, getAddressType } from '../address';
 import { remove0x } from '../utils';
+import { DataCache } from './cache';
 
 export class DataSource {
+  public cache: DataCache;
   public service: BtcAssetsApi;
   public networkType: NetworkType;
 
   constructor(service: BtcAssetsApi, networkType: NetworkType) {
     this.service = service;
     this.networkType = networkType;
+    this.cache = new DataCache();
   }
 
   // Query a UTXO from the service.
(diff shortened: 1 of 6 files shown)


ahonn (approved):

packages/btc/src/transaction/build.ts
```diff
@@ -237,6 +247,8 @@ export class TxBuilder {
         address: props.address,
         targetAmount: _targetAmount,
         allowInsufficient: true,
+        noUtxosCache: true,
+        cacheKey: props.cacheKey,
```
Flouse, line 251: It's a bit confusing to pass a cacheKey while `noUtxosCache=true`.

As my understanding to the implementation,
`collectSatoshi` uses cache by default, only the `cache getter()` does not use cache.

packages/btc/src/query/source.ts
```diff
@@ -117,12 +122,20 @@ export class DataSource {
...
-      only_confirmed: onlyConfirmedUtxos,
-      min_satoshi: minUtxoSatoshi,
+
+    const utxos = await this.cache.optionalCacheUtxos({
+      key: cacheKey,
+      getter: () =>
+        this.getUtxos(address, {
+          only_confirmed: onlyConfirmedUtxos,
+          min_satoshi: minUtxoSatoshi,
+          no_cache: noUtxosCache,
```
Flouse, line 137:
```suggestion
          no_cache: true,
```

packages/btc/src/query/source.ts
```diff
@@ -102,6 +105,8 @@ export class DataSource {
     allowInsufficient?: boolean;
     onlyNonRgbppUtxos?: boolean;
     onlyConfirmedUtxos?: boolean;
+    noUtxosCache?: boolean;
```
Flouse, line 108: maybe the `noUtxosCache` prop could be removed?

packages/btc/src/transaction/build.ts
Flouse: Ok, I get it.

There are actually two layers of cache in the implementation:
- one is sdk's own cache
- The second layer is the assets-api cahce

Flouse: Then it might be better to rename `noUtxosCache` to `noAssetsApiCache`.

Flouse (approved):

Dawn-githup:
https://pudge.explorer.nervos.org/xudt/0xdc50aa08391b9f7831e08a70884e1094dae8984412e52a8c8371ca3bbccc44f5
### Verify that there are 36 cells：

- The develop branch is constructing transaction time：
image

Dawn-githup (approved):

duanyytop:
> https://pudge.explorer.nervos.org/xudt/0xdc50aa08391b9f7831e08a70884e1094dae8984412e52a8c8371ca3bbccc44f5
> 
> ### Verify that there are 36 cells：
> * develop branch build transaction time:：
> 
> image
> * This branch builds transaction time:
> 
> image

Is there still room for improvement in the time cost reduction by the cache? Or is this test case not typical enough?

duanyytop (approved):

packages/btc/src/transaction/build.ts
ShookLyngs: Options has been renamed: https://github.com/ckb-cell/rgbpp-sdk/pull/184/commits/092e3166195a38c064145f14d46f3697919a6e8b

ShookLyngs:
Just renamed the cache-related options to make them look clearer:
- noUtxosCache -> noAssetsApiCache
- cacheKey -> internalCacheKey

ShookLyngs:
> Is there still room for improvement in the time cost reduction by the cache? Or is this test case not typical enough?

The problem can be broken down into two parts:
1. The test case is using the `sendRgbppUtxos()` API for testing, which is indeed not typical enough because this PR aims to improve the construction time for transactions that collect a large amount of inputs during the fee-paying process.
2. The `sendRgbppUtxos()` API can be optimized individually, as it contains time-costly syntax like `for (...) { await fetch() }`, which can be refactored using the `await Promise.all()` API.

PR #246, opened 2024-06-27
refactor(rgbpp-sdk): Support ESM package 

duanyytop:
## Main Changes

- Replace turbo with [tsup](https://tsup.egoist.dev/#what-can-it-bundle)
- Export [CommonJS(cjs)](https://nodejs.org/api/modules.html#modules-commonjs-modules) and [ECMAScript(esm)](https://nodejs.org/api/esm.html) packages for rgbpp sdk
- Use [tsx](https://tsx.is/) instead of `ts-node` to run examples and tests
- Upgrade `ckb-sdk-js` to support ESM
- Clean up the CKB library import packages


ShookLyngs:
I will test if the following changes bring any issues:
- [ ] The `clean` related commands are deleted
- [ ] The `tsconfig.json` and `tsconfig.build.json` config files are deleted
- [ ] The `vitest.config.mts` in btc/service are deleted, and the config in ckb is moved to the repo-root
- [ ] lint-staged in the `package.json` calls npm scripts instead of direct calling fix commands of eslint/prettier

---

@Dawn-githup Can you test if the refactored libs can be read as expected in ESM/CJS projects?

ShookLyngs:
The new commits [4f61ed8](https://github.com/ckb-cell/rgbpp-sdk/pull/246/commits/4f61ed808aa5a77cf94f39f9e39defeb5e262bf7), [6096395](https://github.com/ckb-cell/rgbpp-sdk/pull/246/commits/6096395c9f9eed4ca933c8e1f59245005c80e6bf) and [53fc8ee](https://github.com/ckb-cell/rgbpp-sdk/pull/246/commits/53fc8eec4fb15974f4734de3913e528f4b1d0efc) should have resolved the previous issues. If they occur again, we can update them in new comments to make the entire timeline clearer.

ShookLyngs:
### Discussion: **Should we change the `moduleResolution` to `NodeNext`?**

Previously, when testing `rgbpp` in ESM, an error was reported indicating that subpath imports should always contain file extensions. For example, when importing a JS file, the `.js` extension must be included in the import syntax:

```ts
import { xx } from './a'; // Invalid
import { xx } from './a.js'; // Valid
```

Ref to documentation of Node.js for more: https://nodejs.org/api/esm.html#mandatory-file-extensions

Since all our source files will be packed into a single file when bundling with `tsup`, there should be no `import { xx } from './a'` syntax. The only type of imports left after bundling are package imports for importing packages. To address the issue, all subpath package imports in the packages have been refactored to include the `.js` file extension:

```ts
import { xx } from 'pkg/lib/some'; // From
import { xx } from 'pkg/lib/some.js'; // To
```

However, there is still an issue: our current `tsconfig` settings do not enforce the inclusion of file extensions in imports, and both `tsc` and `tsup` do not handle file extension in the imports. Related discussions:
- https://github.com/nodejs/node/issues/46006
- https://github.com/microsoft/TypeScript/issues/50501

This leads to an incompatibility problem where, if in the future we add new package dependencies, there is nothing to help us check or validate if the new imports are handled correctly:

```ts
import { a } from 'old-pkg/lib/some.js'; // Handled correctly
import { b } from 'new-pkg/lib/other'; // Lacks ".js" extension in the path, but no warning
```

### Possible solutions:
1. Update `"moduleResolution": "NodeNext"` in `tsconfig.json`. This requires all relative imports to contain file extensions, but we would need to refactor all our code to satisfy this requirement.
2. Configure `eslint` to add a rule that requires package imports to include file extension: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/extensions.md.
3. Add "bundle libs" and "execute bundled libs" as a form of integration testing to ensure packages can be executed in both CJS and ESM environments.

Ideas and comments are welcomed.

duanyytop:
None of the three solutions are perfect for me. 

The first two solutions require adding a .js extension to all relative path imports, which is too much of a change and not the recommended practice. 

The third solution seems feasible, but I doubt how many people will encounter this issue. We can pay more attention to the subpath imports during development and code review, and if we indeed miss it and someone reports the problem, it doesn't seem like a big deal to fix it then.

packages/btc/tsconfig.json
```diff
@@ -1,20 +1,19 @@
 {
   "compilerOptions": {
-    "moduleResolution": "Node",
+    "moduleResolution": "Bundler",
```
duanyytop, line 3: The TypeScript document recommends using `"moduleResolution": "Node16"` instead of `"moduleResolution": "Bundler"` for writing a library. 

See: https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html#im-writing-a-library

duanyytop, line 3: The TypeScript document recommends using `"moduleResolution": "Node16"` instead of `"moduleResolution": "Bundler"` for writing a library. 

> In short, "moduleResolution": "bundler" is infectious, allowing code that only works in bundlers to be produced. Likewise, "moduleResolution": "nodenext" is only checking that the output works in Node.js, but in most cases, module code that works in Node.js will work in other runtimes and in bundlers.

See: https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html#considerations-for-bundling-libraries

duanyytop: > Update "moduleResolution": "NodeNext" in tsconfig.json. This requires all relative imports to contain file extensions, but we would need to refactor all our code to satisfy this requirement.

According to the above comment, the current code has added `.js` extension for the relative imports, so we should use `"moduleResolution": "NodeNext"` instead of `"moduleResolution": "Bundler"`.

ShookLyngs: ### My thoughts, at the beginning

According to the TS 5.0+ related documentation, both options are recommended for modern projects. However, after some tests, I think the `Bundler` option is somewhat more reasonable to go with. Though this is not final, so you can continue reading. Any feedback is welcome.

Firstly, I think we have 3 paths to go:
- Specify the packages as `"type": "module"`, and specify `"moduleResolution": "NodeNext"`
- Keep the packages as `"type": "commonjs"`, and specify `"moduleResolution": "NodeNext"`
- Keep the packages as `"type": "commonjs"`, and specify `"moduleResolution": "Bundler"`

And I think the `Bundler` option is more reasonable:
1. We're in fact using `tsup` as the bundler, which transpiles/packs our source code into single CJS/ESM files.
2. The `Bundler` option is an easier option for migration as it prevents a huge refactor of our code.

Let's first look at the difference between `NodeNext` and `Bundler`.

### Difference between `NodeNext` and `Bundler`

While the two options mostly behave the same in the ways we care, and both of them are recommended options for modern projects, yet they do have differences. Refer to this PR for more: https://github.com/microsoft/TypeScript/pull/51669.

The `NodeNext` option requires all relative imports in ESM to include a file extension, meaning all the following import examples should be refactored to contain a file extension `.js` at the end of the path:
```ts
import { a } from './a';      // Invalid
import { a } from './a.js';   // Correct

import { b } from 'pkg/b';    // Invalid
import { b } from 'pkg/b.js'; // Correct
```

The `Bundler` option, however, doesn't treat file extensions as mandatory and will leave the problem to our bundler: `tsup`. While bundling, `tsup` packs all source code into a single JS file, which means in our bundled code, there are no local imports, only the package imports remain:
```ts
import { a } from './a';      // This line doesn't exist in the bundled code
import { b } from 'pkg/b';    // This line exists in the bundled code
```

So with the `Bundler` option, we only need to deal with the package imports:
```ts
import { b } from 'pkg/b';    // Invalid
import { b } from 'pkg/b.js'; // Correct
```

### File extension are just promises

> For more discussions: https://github.com/microsoft/TypeScript/issues/49083

Another thing about the file extension being mandatory is that our bundlers (`tsc` and `tsup`) refuse to transpile the file extension of the imports from `.ts` to `.js` for us while bundling. That is, if we decide to add file extensions to the relative imports in our source code, even if we're importing a `.ts` file, we have to use `.js` like they're some sort of promises:
```ts
import { a } from './a.js';   // There's only an "a.ts" file in the directory
```

However, adding `.js` to package imports feels okay because they are usually just `.js` files:
```ts
import { b } from 'pkg/b.js'; // The "pkg/b.js" file actually exists
```

### The costs of migrating to ESM-first packages

What if we just want to migrate to specifying the `"type": "module"` and `"moduleResolution": "NodeNext"` together, as they are the most recommended settings for ESM-first packages?

Here's a branch that has done most of the refactor work, where all relative imports are refactored to include a file extension, and `"type": "module"` is added to the `package.json` to specify explicitly that the package is an ESM-first library:
- Commit: [[ckb-cell/rgbpp-sdk@fedb0f6]()](https://github.com/ckb-cell/rgbpp-sdk/commit/fedb0f6c4bad9056c330104087d9a357bfc7e9da)
- Snapshot: [`0.0.0-20240709190558`](https://www.npmjs.com/package/rgbpp/v/0.0.0-snap-20240709190558)
- Changes:
  1. Add file extension `.js` to all relative imports.
  2. Add `"type": "module"` to `package.json` to tag explicitly that this is an ESM-first package.
  3. Fix some type issues or weird incompatibilities found in the `rgbpp-sdk/ckb` lib to prevent bundling failures.
- Test results:
  - Tested in https://github.com/ShookLyngs/test-rgbpp-pure-esm in both ESM/CJS.
  - Tested in a CLI project to bundle it into a single executable application.

### Use `NodeNext` without migration

While testing everything, I find that we could specify the `NodeNext` option as the module resolution and NOT set `"type": "module"` in the `package.json`, which means by default our packages are just treated as CJS. With this combination, the TSC actually reports no error about our relative imports, even if they have no file extension included.

The problems are:
1. I'm not sure if this is a valid thing to do, or if it is simply an unresolved bug. Any documentation about this?
2. If we're using this combination, what's the difference from using the `Bundler` option? As they both don't give warnings from the TSC and IDE, we still need to validate the packages after bundling them.

### Conclusion

After everything, these are my thoughts about the combinations:

- Specify the packages as `"type": "module"`, and specify `"moduleResolution": "NodeNext"`
  - We have to refactor most of the code, and the file extensions in the imports are not what they look like.
  - If we decide to overcome the downsides, this is the recommended choice by the community.
- Keep the packages as `"type": "commonjs"`, and specify `"moduleResolution": "NodeNext"`
  - Behaves the same as `Bundler`. Since we're using `tsup`, I think this other option is more appropriate.
  - I'm not sure if this combination is a bug of TSC or if it is simply a valid choice.
- Keep the packages as `"type": "commonjs"`, and specify `"moduleResolution": "Bundler"`
  - The bundled code is not promised to work at runtime because file extensions for package imports are not mandatory, therefore lacking the process to validate the source code.
  - If we treat `tsup` as a bundler, we can use this option.

duanyytop: Have you tried with the `"module": "node16"` and `"moduleResolution": "nodenext"` of the `tsconfig.json`.

See https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html#im-writing-a-library

- module: "node16". When a codebase is compatible with Node.js’s module system, it almost always works in bundlers as well. If you’re using a third-party emitter to emit ESM outputs, ensure that you set "type": "module" in your package.json so TypeScript checks your code as ESM, which uses a stricter module resolution algorithm in Node.js than CommonJS does. As an example, let’s look at what would happen if a library were to compile with "moduleResolution": "bundler":

```
export * from "./utils";
```

Assuming ./utils.ts (or ./utils/index.ts) exists, a bundler would be fine with this code, so "moduleResolution": "bundler" doesn’t complain. Compiled with "module": "esnext", the output JavaScript for this export statement will look exactly the same as the input. If that JavaScript were published to npm, it would be usable by projects that use a bundler, but it would cause an error when run in Node.js:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../node_modules/dependency/utils' imported from .../node_modules/dependency/index.js
Did you mean to import ./utils.js?
```

ShookLyngs: Yes, I've tried it, a TS error will be thrown if we don't use them in pairs (that's why I didn't mention it):

```
error TS5110: Option 'module' must be set to 'NodeNext' when option 'moduleResolution' is set to 'NodeNext'
```

FYI, Node16 and NodeNext are currently identical, according to the documentation: https://www.typescriptlang.org/docs/handbook/modules/reference.html#node16-nodenext.

ShookLyngs (approved):
I'm giving a like to the PR because I think the current implementation works by keeping `"type": "commonjs"` in the `package.json`, and specify `"moduleResolution": "Bundler"` in the `tsconfig.json`. Also, the implementation includes the least changes to the codease.

duanyytop:
I have tested in JoyID, Next.js, and Node.js scripts with `rgbpp-sdk`(0.0.0-snap-20240801024204) whose packages both support ESM and CommonJS, and they all worked well.

- JoyID: https://joyid-app-git-rgbpp-test-nervina.vercel.app/
- Next.js project: https://github.com/duanyytop/node-esm-test/tree/master/apps/next-rgbpp
- Node scripts: https://github.com/duanyytop/node-esm-test/blob/master/apps/scripts/lock-seal-onchain.mjs

cc @ShookLyngs @Flouse

Flouse (dismissed):

apps/service/package.json
```diff
@@ -28,7 +28,7 @@
     "convert-keys": "^1.3.4",
     "json-rpc-2.0": "^1.7.0",
     "reflect-metadata": "^0.2.0",
-    "rgbpp": "workspace:*",
+    "rgbpp": "^0.4.0",
```
Flouse, line 31:
```suggestion
    "rgbpp": "^0.5.0",
```

Flouse (approved):

ShookLyngs (approved):

PR #256, opened 2024-07-18
fix(rgbpp-sdk/ckb): Update spore capacity if it is insufficient

duanyytop:
## Changes

- Increase spore cell capacity with 95CKB if the capacity is insufficient

### DIFF
--- a/packages/ckb/src/utils/ckb-tx.ts
+++ b/packages/ckb/src/utils/ckb-tx.ts
@@ -162,6 +162,11 @@ export const isSporeCapacitySufficient = (sporeCell: IndexerCell) => {
   return capacity - occupiedCapacity > BigInt(BTC_TIME_CELL_INCREASED_SIZE) * CKB_UNIT;
 };
 
+// If the spore cell capacity is insufficient, increase it by BTC_TIME_CELL_INCREASED_SIZE
+export const increaseSporeCapacity = (currentSporeCapacity: Hex): Hex => {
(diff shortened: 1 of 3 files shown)


packages/ckb/src/utils/ckb-tx.ts
```diff
@@ -162,6 +162,11 @@ export const isSporeCapacitySufficient = (sporeCell: IndexerCell) => {
   return capacity - occupiedCapacity > BigInt(BTC_TIME_CELL_INCREASED_SIZE) * CKB_UNIT;
 };
 
+// If the spore cell capacity is insufficient, increase it by BTC_TIME_CELL_INCREASED_SIZE
+export const increaseSporeCapacity = (currentSporeCapacity: Hex): Hex => {
+  return `0x${(BigInt(currentSporeCapacity) + BigInt(BTC_TIME_CELL_INCREASED_SIZE) * CKB_UNIT).toString(16)}`;
```
Flouse, line 167: LGTM.

another way is `return calculateRgbppSporeCellCapacity(..., reserveMoreCkb = true)`

Flouse (approved):

packages/ckb/src/utils/ckb-tx.ts
duanyytop: They are different.

Assume the occupied capacity is 200CKB, the `calculateRgbppSporeCellCapacity` will return `200 + 1 + 95` and the `increaseSporeCapacity` will return `200.99876543 + 95`. 

`currentSporeCapacity` will be bigger than 200CKB and smaller than 201CKB(some CKB has already been used for transaction fee), I think the `increaseSporeCapacity` will be better.

PR #270, opened 2024-08-06
feat(rgbpp): support for batch transferring of RGBPP XUDT assets

ShookLyngs:
## Changes
- Add `buildRgbppTransferAllTxs()` API to generate one or more BTC/CKB transaction groups for transferring the entire amount of a specific type of RGBPP XUDT asset from one or more BTC addresses to a recipient
- Add `sendRgbppTxGroups()` API for sending BTC/CKB transaction groups to the `BtcAssetsApi`

## TODOs
- [x] Base coding
- [x] Testing
  - [x] Prepare mocked tests @ShookLyngs 
  - [x] Functional testing @Dawn-githup 
- [x] Update docs (README of rgbpp)
- [ ] ~~Update examples? (no place for the example)~~


ShookLyngs:
Mocked a test for the `transfer-all-of-an-rgbpp-xudt-asset` feature with @Dawn-githup:
- Transfer all amount of a specific RGBPP XUDT asset `XTT`, from Alice to Bob
- Alice holds 41 L2 cells bound to the same L1 output, the total amount is `410 XTT`
- Alice holds another 50 L2 cells, each bound to an individual L1 output, the total amount is `500 XTT`

The expected result:
- Alice transfers `500 XTT` to Bob in 2 transactions because each CKB can contain up to 40 RGBPP cells
- The 41 cells with total worth of `410 XTT` should be excluded during the transfer process, because too much L2 cells bound to one L1 output
- The L1 fees should be paid with the L1 RGBPP outputs because the fee rate is low

The actual transaction groups generated from the process:
- [0da44932270292fd3a4165f1f7ab81abf69b951e1e7d3b5c012e00a291c6b222](https://pudge.explorer.nervos.org/transaction/0da44932270292fd3a4165f1f7ab81abf69b951e1e7d3b5c012e00a291c6b222)
- [25414c272e974192cafba60ccb07575c841b26f7e4da5d89e853f4bf75db5add](https://pudge.explorer.nervos.org/transaction/25414c272e974192cafba60ccb07575c841b26f7e4da5d89e853f4bf75db5add)

packages/rgbpp/src/rgbpp/utils/transaction.ts
```diff
@@ -0,0 +1,61 @@
...
+    txId?: string;
+    error?: Error | unknown;
+  };
+  ckb: {
+    sent: boolean;
+    error?: Error | unknown;
+  };
+}
+
+export async function sendRgbppTxGroups(props: {
```
ShookLyngs, line 21: Do you think the `sendRgbppTxGroups` method should be exposed to users? @Flouse @duanyytop

.github/workflows/test.yaml
```diff
@@ -52,6 +52,8 @@ jobs:
       - name: Run tests for packages
         run: pnpm run test:packages
         env:
+          VITE_CKB_NODE_URL: https://mainnet.ckb.dev/rpc
+          VITE_CKB_INDEXER_URL: https://mainnet.ckb.dev/indexer
```
Flouse, line 56: Why use mainnet RPC here?

ShookLyngs: Fixed at 28526453

packages/rgbpp/src/rgbpp/utils/transaction.ts
```diff
@@ -0,0 +1,61 @@
...
+    const result: SentRgbppTxGroup = {
+      ckb: {
+        sent: false,
+      },
+      btc: {
+        sent: false,
+      },
+    };
+    try {
+      // TODO: add retry logic
```
duanyytop, line 37: Is there any logic to be implemented?

packages/rgbpp/tests/shared/account.ts
```diff
@@ -0,0 +1,80 @@
+import {
```
duanyytop, line 1: I suggest this file should be moved to `rgbpp-sdk/btc` and examples can reuse it

packages/rgbpp/tests/shared/file.ts
```diff
@@ -0,0 +1,24 @@
+import { existsSync, mkdirSync, writeFileSync } from 'node:fs';
+import { resolve, join } from 'node:path';
+
+export function saveJson(props: { filename: string; json: object; path?: string }) {
```
duanyytop, line 4: I cannot find any usage of this function

packages/rgbpp/src/rgbpp/utils/transaction.ts
```diff
@@ -0,0 +1,61 @@
...
+      const sent = await props.btcService.sendBtcTransaction(group.btcTxHex);
+      result.btc.txId = sent.txid;
+      result.btc.sent = true;
+    } catch (e) {
+      result.btc.sent = false;
+      result.btc.error = e;
+    }
+    if (result.btc.sent) {
+      try {
+        await props.btcService.sendRgbppCkbTransaction({
```
duanyytop, line 47: It is recommended to place `sendRgbppCkbTransaction` right after `sendBtcTransaction`, so there will be no need to maintain a status list of whether the transactions were sent successfully.

packages/rgbpp/src/rgbpp/xudt/btc-transfer-all.ts
```diff
@@ -0,0 +1,219 @@
...
+    },
+    new Map<string, Utxo>(),
+  );
+
+  // Get each L1 UTXO's corresponding L2 Cells
+  // If a UTXO has more L2 Cells than maxRgbppCellsPerCkbTx, the UTXO/Cells are invalid.
+  // Invalid L1 UTXO and its corresponding L2 Cells are excluded from the transfer process.
+  const rgbppGroups: { id: string; cells: number }[] = [];
+  await Promise.all(
+    [...utxoIds].map(async (utxoId) => {
```
duanyytop, line 77: Is using `utxoIds` instead of `[...utxoIds]` more simple?

Flouse:
> * The L1 fees should be paid with the L1 RGBPP outputs because the fee rate is low

Have you tested the case where the transaction requires additional inputs to pay tx fee?
@ShookLyngs @Dawn-githup

packages/rgbpp/src/rgbpp/xudt/btc-transfer-all.ts
```diff
@@ -0,0 +1,219 @@
...
+  // 2. Each CKB_VTX can only include (<= maxRgbppCellsPerCkbTx) amount of Cells
+  const boundCells = rgbppGroups.map((group) => group.cells);
+  const groupedByInputs = groupNumbersBySum(boundCells, maxRgbppCellsPerCkbTx);
+  const groupedAssetGroups = mapGroupsByIndices(groupedByInputs.indices, (index) => rgbppGroups[index]!);
+
+  // Construct transaction groups
+  const summarizer = new AssetSummarizer();
+  const usedBtcUtxos: BaseOutput[] = [];
+  const transactionGroups: RgbppTransferAllTxGroup[] = [];
+  for (const assetGroups of groupedAssetGroups) {
```
duanyytop, line 113: There are async functions in the loop. If we don't use `await`, the `assetGroups` might get messed up.

packages/rgbpp/src/rgbpp/utils/transaction.ts
duanyytop: I think it is needed for developers

Dawn-githup:
> > * The L1 fees should be paid with the L1 RGBPP outputs because the fee rate is low
> 
> Have you tested the case where the BTC transaction requires additional inputs to pay tx fee? @ShookLyngs @Dawn-githup

When the fee rate is higher, utxo will be collected from the from address to pay the fee.

packages/rgbpp/tests/shared/file.ts
ShookLyngs: It was used during test mocking, I feel like it's still needed in the future.

packages/rgbpp/src/rgbpp/utils/transaction.ts
ShookLyngs: The retry logic is not implemented yet. I was not sure about the feature, so I left a TODO comment.

packages/rgbpp/tests/shared/account.ts
ShookLyngs: Maybe we should, a concern is that we may not handle P2TR well, because it can be complicated.

packages/rgbpp/src/rgbpp/utils/transaction.ts
ShookLyngs: Do you mean like this?
```ts
try {
  // TODO: add retry logic
  const sent = await props.btcService.sendBtcTransaction(group.btcTxHex);
  result.btc.txId = sent.txid;
  result.btc.sent = true;
  
  try {
    await props.btcService.sendRgbppCkbTransaction({
      btc_txid: result.btc.txId!,
      ckb_virtual_result: group.ckbVirtualTxResult,
    });
    result.ckb.sent = true;
  } catch (e) {
    result.ckb.sent = false;
    result.ckb.error = e;
  }
} catch (e) {
  result.btc.sent = false;
  result.btc.error = e;
}
```

packages/rgbpp/src/rgbpp/xudt/btc-transfer-all.ts
ShookLyngs: The `for await...of` syntax is for when going through an async iterable object, since the `groupedAssetGroups` variable is just a regular object, I think it's fine without the `await` syntax. Ref to: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of

packages/rgbpp/src/rgbpp/xudt/btc-transfer-all.ts
ShookLyngs: You're right, I'm not sure why I wrote a clone here, I'll check and fix it.

packages/rgbpp/src/rgbpp/utils/transaction.ts
duanyytop: Yes. 

I think `result.btc.sent` is not needed, if the BTC `txId` or `result.btc.error` is not null, the BTC transaction must be committed successfully, and the `result.ckb.sent` is not needed, either.

packages/rgbpp/tests/shared/file.ts
duanyytop: If it is needed, adding it again will be more reasonable.

packages/rgbpp/src/rgbpp/xudt/btc-transfer-all.ts
ShookLyngs: I double-checked the code, turns out this part is expected because `utxoIds` is a `Set<string>` object and does not have a `.map()` method, which is why I used the `[...utxoIds]` syntax.

packages/rgbpp/tests/shared/file.ts
ShookLyngs: Updated at 0f8f523

packages/rgbpp/src/rgbpp/utils/transaction.ts
ShookLyngs: The "retry" prop is removed for now at 37074c4

packages/rgbpp/src/rgbpp/utils/transaction.ts
ShookLyngs: The `result.x.sent` props have removed, updated at 37074c4. Tho I'm not wrapping try-catch blocks due to readability reasons.

Flouse:
### Discussion

- [ ] Update apps/service in a separate PR?

packages/btc/src/utils.ts
```diff
@@ -77,6 +79,44 @@ export function transactionToHex(tx: bitcoin.Transaction, withWitness?: boolean)
...
+  const parts = utxoId.split(':');
+  const txid = parts[0];
+  const vout = parts[1] ? parseInt(parts[1]) : undefined;
+  if (
+    !txid ||
+    txid.startsWith('0x') ||
+    txid.length !== 64 ||
+    typeof vout !== 'number' ||
+    isNaN(vout) ||
+    vout > 0xffffffff
```
Flouse, line 109:
```suggestion
    vout > 0xffffffff ||
    vout < 0
```

packages/rgbpp/README.md
```diff
@@ -49,4 +49,49 @@ const { txid: btcTxId } = await btcService.sendBtcTransaction(btcTx.toHex());
...
+  btc: {
+    assetAddresses,
+    fromAddress,
+    toAddress,
+    dataSource,
+    feeRate,
+    pubkeyMap,
+  },
+  isMainnet,
+});
```
Flouse, line 75: ### Discussion
Add the comments of https://github.com/ckb-cell/rgbpp-sdk/blob/b1865c2ac35ef3dec68afbd81e0736cbff80edd0/packages/rgbpp/src/rgbpp/types/xudt.ts#L45-L76 to this README, to provide a clear reference for developers working with `buildRgbppTransferAllTxs`, especially when dealing with `assetAddresses`, `fromAddress`, `toAddress` arguments.

packages/rgbpp/src/rgbpp/summary/asset-summarizer.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export interface AssetGroupSummary {
+  utxoId: string;
+  cellIds: string[];
+  assets: Record<string, AssetSummary>;
+}
+
+export interface TransactionGroupSummary {
+  utxos: number;
+  cells: number;
```
Flouse, line 19: ### Suggestion
  utxoCount: number;
  cellCount: number;

packages/rgbpp/src/rgbpp/summary/asset-summarizer.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export class AssetSummarizer {
+  groups: AssetGroupSummary[] = [];
+
+  constructor() {}
+
+  addGroup(utxo: Utxo, cells: Cell[]): AssetGroupSummary {
+    const utxoId = encodeUtxoId(utxo.txid, utxo.vout);
+    const assets: Record<string, Omit<AssetSummary, 'xudtTypeArgs'>> = {};
```
Flouse, line 32: The `AssetSummary` interface does NOT contain a `xudtTypeArgs` property.
Maybe the `Omit` type is unnecessary in this context?

packages/rgbpp/src/rgbpp/summary/asset-summarizer.ts
```diff
@@ -0,0 +1,98 @@
...
+  amount: bigint;
+  utxos: number;
+  cells: number;
+}
+
+export interface AssetGroupSummary {
+  utxoId: string;
+  cellIds: string[];
+  assets: Record<string, AssetSummary>;
+}
```
Flouse, line 15:
```suggestion
export interface AssetGroupSummary {
  utxoId: string;
  cellIds: string[];
  // The key of the assets record is the `xudtTypeArgs` (the unique identifier for the asset type)
  assets: Record<string, AssetSummary>;
}
```

packages/rgbpp/src/rgbpp/summary/asset-summarizer.ts
```diff
@@ -0,0 +1,98 @@
+import { Cell } from '@ckb-lumos/base';
+import { Utxo, encodeUtxoId } from '@rgbpp-sdk/btc';
+import { leToU128, encodeCellId } from '@rgbpp-sdk/ckb';
+
+export interface AssetSummary {
+  amount: bigint;
+  utxos: number;
```
Flouse, line 7: When is `AssetSummary.utxos` field used? Is it necessary?
Because each `AssetSummary` represents the assets bound to a single BTC UTXO,
when an `AssetSummary` is created, `utxoCount` is always be `1`, right?

Update:
ref: https://github.com/ckb-cell/rgbpp-sdk/blob/b1865c2ac35ef3dec68afbd81e0736cbff80edd0/packages/rgbpp/src/rgbpp/summary/asset-summarizer.ts#L81

packages/rgbpp/src/rgbpp/summary/asset-summarizer.ts
```diff
@@ -0,0 +1,98 @@
...
+    this.groups.push(result);
+    return result;
+  }
+
+  addGroups(groups: { utxo: Utxo; cells: Cell[] }[]): TransactionGroupSummary {
+    const groupResults = groups.map((group) => this.addGroup(group.utxo, group.cells));
+    return this.summarizeGroups(groupResults);
+  }
+
+  summarizeGroups(groups?: AssetGroupSummary[]): TransactionGroupSummary {
```
Flouse, line 66:
```suggestion
  /**
   * Summarizes asset information across multiple AssetGroupSummary objects.
   * If no groups are provided, it summarizes the internally stored groups.
   * 
   * @param groups - Optional array of AssetGroupSummary objects to summarize. 
   * @returns TransactionGroupSummary - A summary of all assets across the specified groups,
   *          including total UTXO and cell counts, UTXO and cell IDs, and aggregated asset information.
   */
  summarizeGroups(groups?: AssetGroupSummary[]): TransactionGroupSummary {
```

packages/rgbpp/src/rgbpp/summary/asset-summarizer.ts
```diff
@@ -0,0 +1,98 @@
...
+          if (result[xudtTypeArgs] === undefined) {
+            result[xudtTypeArgs] = {
+              utxos: 0,
+              cells: 0,
+              amount: 0n,
+            };
+          }
+
+          result[xudtTypeArgs]!.utxos += summary.assets[xudtTypeArgs]!.utxos;
+          result[xudtTypeArgs]!.cells += summary.assets[xudtTypeArgs]!.cells;
```
Flouse, line 82: It seems that `AssetSummary.{utxos, cells}` are unused.

packages/rgbpp/src/rgbpp/summary/asset-summarizer.ts
ShookLyngs: > When is AssetSummary.utxos field used? Is it necessary?

The `summary` object in the result is also for users to review or display on an app. So, even though some props are not used in our code, they can still help users identify which UTXOs/Cells are included/excluded during the process, and how many of them are included/excluded.

> Because each AssetSummary represents the assets bound to a single BTC UTXO, when an AssetSummary is created, utxoCount is always be 1, right?

Each `AssetSummary` represents the summary of a single RGBPP XUDT Asset instead of a single UTXO, therefore the `utxoCount` can be greater than `1`.

packages/btc/src/utils.ts
ShookLyngs: Fixed at 75840f8

packages/rgbpp/src/rgbpp/summary/asset-summarizer.ts
ShookLyngs: Updated at 1440fb8

packages/rgbpp/src/rgbpp/summary/asset-summarizer.ts
ShookLyngs: Fixed at 1440fb8

packages/rgbpp/src/rgbpp/summary/asset-summarizer.ts
ShookLyngs: Updated at 1440fb8

packages/rgbpp/src/rgbpp/summary/asset-summarizer.ts
ShookLyngs: Updated at 1440fb8

ShookLyngs:
@duanyytop and I had a few discussions about the PR:

1. The `.env` file for the rgbpp lib is redundant, as all tests in the rgbpp lib are mocked. Can it be removed?
2. Should we keep the rgbpp lib as simple as possible, instead of adding a lot of utilities to it?
3. Developers may need an example for the transfer-all feature when coding.

My thoughts:

> The `.env` file for the rgbpp lib is redundant, as all tests in the rgbpp lib are mocked. Can it be removed?

The btc/ckb lib contain some methods that are not mockable, meaning that when calling the `buildRgbppTransferAllTxs` API, actual network requests must be sent. So, while the `.env` file can be removed, we still need to provide a default environment so the tests can run as expected. I'm not sure if it's worth it.

I think @duanyytop is concerned about redundant test cases, as there are too many "same pieces of code" in the examples and integration tests. If we add more tests to the rgbpp lib alone, that would increase redundancy. 

Personally, I believe the unit tests for the rgbpp library are different from the examples and integration tests because the unit tests allow more room for unexpected cases. However, writing tests for the `buildRgbppTransferAllTxs` API is challenging and requires a lot of preparation. Maybe @Dawn-github can write integration tests for the feature, and then we can remove the mocked tests from the rgbpp lib, making the `.env` file unnecessary.

> Should we keep the rgbpp library as simple as possible, instead of adding a lot of utilities?

Agreed, so I moved most of the utilities from the rgbpp library to the btc/ckb libs at [c492cf3](https://github.com/ckb-cell/rgbpp-sdk/pull/270/commits/c492cf3fc95507b6db334f9f0c1197fe00302c20).

> Developers may need an example for the transfer-all feature when writing.

Yes, but it would be hard to write a set of examples just to prepare the assets. If we provide only one example for users who already have many assets, it would be much easier. We may add such example in `examples/rgbpp/xudt/btc-transfer-all/1-btc-transfer-all.ts`.

packages/rgbpp/README.md
ShookLyngs: Updated at c7f37a3

packages/rgbpp/tests/shared/account.ts
ShookLyngs: After reviewing the current codebase of the packages/examples, I think this is not a good timing to do this. Examples and tests have different needs for the feature, and it's a big change to refactor them all.

packages/rgbpp/src/rgbpp/summary/asset-summarizer.ts
Flouse: I see.

packages/rgbpp/src/rgbpp/summary/asset-summarizer.ts
Flouse: explained by https://github.com/ckb-cell/rgbpp-sdk/pull/270#discussion_r1712964066

packages/rgbpp/src/rgbpp/xudt/btc-transfer-all.ts
```diff
@@ -0,0 +1,234 @@
...
+        if (cells) {
+          cellsMap.set(utxoId, cells);
+        }
+        const utxo = utxoMap.get(utxoId);
+        if (!utxo || !cells || cells.length > maxRgbppCellsPerCkbTx) {
+          invalidUtxoIds.add(utxoId);
+          return;
+        }
+        cells.forEach((cell) => {
+          const cellId = encodeCellId(cell.outPoint!.txHash, cell.outPoint!.index);
```
Flouse, line 104: ### Discussion
If a BTC UTXO is bound to a DOB cell, maybe it should be added to invalidUtxoIds too?

## Related Test Case
- a BTC UTXO is bound to multi type_scripts (xUDT + DOB)

Flouse (approved):

Dawn-githup:
**Regarding the transfer-all example, we are currently facing an issue:** if we use [this file](https://github.com/ckb-cell/rgbpp-sdk/blob/develop/examples/rgbpp/shared/btc-account.ts) to create a BTC account and sign transactions, with `createBtcAccount` it won’t support mixed-type addresses when using `.env`.

- For example, if we write it as "A wants to transfer all of their xudt1 to B," then the fee and change can go to A, and there's no need for multiple accounts. Would such simple support be sufficient for the example?

@Flouse @ShookLyngs

packages/rgbpp/src/rgbpp/xudt/btc-transfer-all.ts
Dawn-githup: > If a BTC UTXO is bound to a DOB cell, maybe it should be added to invalidUtxoIds too?

I think there is no need to mark as invalid (add to invalidUtxoIds list;

- rgb-sdk will not handle this kind of asset and throw an exception;

- excluded.assets displays > 40(RGBPP_TX_INPUTS_MAX_LENGTH) assets

ShookLyngs: I tested this part with @Dawn-github. If a UTXO is bound to any non-supported type cells, an error will be thrown from the ckb lib. Essentially, if the `codeHash` and `hashType` of the cell's type script do not match the expected values, the type is not supported.
```
Error: The BTC UTXO must not be bound to xUDT and other type cells at the same time
```

I think I need to add a check here.

ShookLyngs:
> **Regarding the transfer-all example, we are currently facing an issue:** if we use [this file](https://github.com/ckb-cell/rgbpp-sdk/blob/develop/examples/rgbpp/shared/btc-account.ts) to create a BTC account and sign transactions, with `createBtcAccount` it won’t support mixed-type addresses when using `.env`.

I think it depends on how complicated the examples need to be. If we're writing a simple example that transfers all XUDT1 from A to B, then a single account is enough. However, if we're writing an example that transfers all XUDT1 from [A, B] to C, then we'll need at least 2 accounts for signing.

packages/rgbpp/src/rgbpp/xudt/btc-transfer-all.ts
ShookLyngs: Fixed at cd644176

Flouse (approved):

Dawn-githup (approved):

Dawn-githup:
> It's best to draft a TODO list of the examples first. cc @Flouse @Dawn-githup

- This is a simple example using a single account transfer-all

https://github.com/ckb-cell/rgbpp-sdk/commit/10f045774d705adbf44b1dcf3bf3ce34866a433b

Flouse:
> * This is a simple example using a single account transfer-all
> 
> [10f0457](https://github.com/ckb-cell/rgbpp-sdk/commit/10f045774d705adbf44b1dcf3bf3ce34866a433b)

ok, please add an introduction to https://github.com/ckb-cell/rgbpp-sdk/blob/develop/examples/rgbpp/README.md to explain the use case of this example later.

PR #278, opened 2024-08-13
feat(rgbpp-sdk-service): support `generate_rgbpp_transfer_all_txs` API

ShookLyngs:
## Changes
- Support `generate_rgbpp_transfer_all_txs` in the rgbpp-sdk-service for batch transferring of RGB++ xUDT assets, previously implemented at https://github.com/ckb-cell/rgbpp-sdk/pull/270
- Refactor `toSnakeCase()` and `toCamelCase()` to provide clearer return types
- Fix `"module": "NodeNext"` related import path issues
    - The `src/*` path shortcut does not work
    - Relative imports must end with `.js`
- Add rgbpp-sdk-service tests to the `test.yml` workflow

### Case transformation issue

The `toSnakeCase()` util transforms all object keys to snake_case, e.g. `snakeCase` to `snake_case`. But when transforming the `TransactionGroupSummary` object, an issue was found that if the object key is a hex string, it will also be transformed as snake_case, e.g. `0xe1e` to `0_xe_1e`. This is unexpected.

In the current implementation of the `generate_rgbpp_transfer_all_txs` API, I wrapped the result with an util `ensureSafeJson()`, which detects if the object key starts with `0_x`, then all underscore marks in the key will be removed, reverting it to a regular hex string, e.g. `0_xe_1e` to `0xe1e`.

```ts
// XXX: Remove underscores from hex strings due to toSnakeCase() transformation issue
if (key.startsWith('0_x')) {
  key = key.replaceAll('_', '');
}
```

### BigInt to String issue

The requests in the `rgbpp-sdk-service` returns JSON as response. An JSON-related issue was found that it doesn't serialize BIgInt values. When an object that contains such type of value is passed to `JSON.stringify()`, the function throws an error:
```
TypeError: Do not know how to serialize a BigInt
```

Currently, I just transform all BIgInt values to hex strings:
```ts
// XXX: Convert BigInt to hex string for JSON.stringify() compatibility
if (typeof value === 'bigint') {
  obj[key] = `0x${value.toString(16)}`;
} else {
  obj[key] = value;
}
```


duanyytop:
> an issue was found that if the object key is a hex string, it will also be transformed as snake_case, e.g. 0xe1e to 0_xe_1e. This is unexpected.

Why does a key with 0x-hex-string appear?

duanyytop:
> Relative imports must end with .js

May I ask why the .js suffix is needed? It worked fine before without the suffix.

apps/service/src/utils/json.ts
```diff
@@ -0,0 +1,28 @@
+import isPlainObject from 'lodash/isPlainObject.js';
+
+export function ensureSafeJson<Input extends object, Output = Input>(json: Input): Output {
```
duanyytop, line 3: We should avoid BigInt data in the parameter instead of converting BigInt to hex-string in the JSON stringify function.

apps/service/src/utils/case.ts
```diff
@@ -0,0 +1,23 @@
+import { SnakeCasedPropertiesDeep, CamelCasedPropertiesDeep } from 'type-fest';
+import { toSnake, toCamel } from 'convert-keys';
+
+export type SnakeCased<T> = SnakeCasedPropertiesDeep<T>;
+export type CamelCased<T> = CamelCasedPropertiesDeep<T>;
+
+export const toSnakeCase = <T>(obj: T): SnakeCased<T> | null => {
```
duanyytop, line 7: Maybe we can use this package https://www.npmjs.com/package/camelcase-keys to convert to snake case with deep mode and it is more simple.

apps/service/src/utils/case.ts
```diff
@@ -0,0 +1,23 @@
...
+  try {
+    return toSnake(obj);
+  } catch (error) {
+    console.error(error);
+  }
+  return null;
+};
+
+export const toCamelCase = <T>(obj: T): CamelCased<T> | null => {
```
duanyytop, line 16: The `toCamelcase` of `ckb` lib can convert data to camel case with deep mode and I can rename the function to `toCamelCase`.

apps/service/src/utils/json.ts
Flouse: https://github.com/search?q=repo%3Ackb-cell%2Frgbpp-sdk%20%3A%20bigint%3B&type=code shows that `bigint` has been used in `packages/ckb/src/types`.
So, maybe this `ensureSafeJson` function is necessary in the current stage？

Flouse (approved):
LGTM

apps/service/src/utils/json.ts
duanyytop: Yes. However, not all types will be converted to JSON strings. If we add the `ensureSafeJson` function, we should add tests to ensure it works fine.

apps/service/package.json
```diff
@@ -27,8 +28,9 @@
     "@nestjs/platform-fastify": "^10.3.9",
     "convert-keys": "^1.3.4",
     "json-rpc-2.0": "^1.7.0",
+    "lodash": "^4.17.21",
     "reflect-metadata": "^0.2.0",
-    "rgbpp": "^0.5.0",
+    "rgbpp": "0.0.0-snap-20240813134030",
```
ShookLyngs, line 33: > May I ask why the `.js` suffix is needed? It worked fine on my local computer without `type: module` and the `.js` suffix.

@duanyytop, it was due to the upgrade of `rgbpp`. It works fine with version `0.5.0`, but the following error is thrown when upgraded to `0.0.0-snap-20240813134030` (a relatively new version of the snapshots). Do you have a clue on how to resolve the issue without specifying `"type": "module"` in the package.json?

```
/Users/root/Development/projects/rgbpp-sdk/node_modules/.pnpm/@rgbpp-sdk+ckb@0.0.0-snap-20240813134030_@ckb-lumos+lumos@0.22.0-next.5_lodash@4.17.21/node_modules/@rgbpp-sdk/ckb/dist/index.js:469
var import_ckb_sdk_core = __toESM(require("@nervosnetwork/ckb-sdk-core"));
                                  ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/root/Development/projects/rgbpp-sdk/node_modules/.pnpm/@nervosnetwork+ckb-sdk-core@0.109.2/node_modules/@nervosnetwork/ckb-sdk-core/lib/index.js from /Users/root/Development/projects/rgbpp-sdk/node_modules/.pnpm/@rgbpp-sdk+ckb@0.0.0-snap-20240813134030_@ckb-lumos+lumos@0.22.0-next.5_lodash@4.17.21/node_modules/@rgbpp-sdk/ckb/dist/index.js not supported.
/Users/root/Development/projects/rgbpp-sdk/node_modules/.pnpm/@nervosnetwork+ckb-sdk-core@0.109.2/node_modules/@nervosnetwork/ckb-sdk-core/lib/index.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead either rename /Users/root/Development/projects/rgbpp-sdk/node_modules/.pnpm/@nervosnetwork+ckb-sdk-core@0.109.2/node_modules/@nervosnetwork/ckb-sdk-core/lib/index.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /Users/root/Development/projects/rgbpp-sdk/node_modules/.pnpm/@nervosnetwork+ckb-sdk-core@0.109.2/node_modules/@nervosnetwork/ckb-sdk-core/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

apps/service/src/rgbpp/types.ts
```diff
@@ -41,3 +43,41 @@ export interface RgbppCkbTxHashReq {
...
+    // note that you must specify a pubkey for each P2TR address in assetAddresses/fromAddress
+    pubkeyMap?: AddressToPubkeyMap;
+    // The BTC address to return change satoshi, default value is fromAddress
+    changeAddress?: string;
+    // The fee rate of the BTC transactions, will use the fastest fee rate if not specified
+    feeRate?: number;
+  };
+}
+
+export interface RgbppTransferAllRes extends Omit<RgbppTransferAllTxsResult, 'transactions'> {
```
ShookLyngs, line 75: > Why does a key with 0x-hex-string appear?

@duanyytop, the hex keys are in the `RgbppTransferAllTxsResult` type for the summary info of the XUDT assets: https://github.com/ckb-cell/rgbpp-sdk/blob/develop/packages/rgbpp/src/rgbpp/summary/asset-summarizer.ts#L20

duanyytop: I see. LGTM

apps/service/src/utils/json.ts
ShookLyngs: Sure, I can add some tests for it. Bold question: Is it possible to refactor all usages of the native `bigint` to use `BI` and hex strings instead? Would this be considered a breaking change, or is it acceptable to refactor it partially?

duanyytop: Only affects JSON stringify for BigInt and I don't think it is necessary to refactor.

apps/service/src/utils/case.ts
ShookLyngs: I think you mean: https://github.com/bendrucker/snakecase-keys

This is actually a good idea, not because the type the package provides, but because it provides a `exclude` option for excluding strings from being transformed. If everything goes well, this package should help resolve the `0xe1e` to `0_xe_1e` issue in another way. I'll give it a try.

duanyytop: Sorry, https://github.com/bendrucker/snakecase-keys is right.

apps/service/src/utils/case.ts
ShookLyngs: I think using third-party packages for the feature is acceptable. If you're updating it, the return type of the `toCamelcase` function in the ckb lib could be refactored to be more accurate, using the `CamelCased<T>` type instead of `T` as the return type.

apps/service/package.json
duanyytop: I retried on my local computer, `.js` suffix is required.

duanyytop (approved):

apps/service/src/utils/case.ts
ShookLyngs: Updated at 27bb01e

duanyytop (approved):

Flouse (approved):

PR #292, opened 2025-01-06
Upgrade dependencies

duanyytop:
Try to fix the errors: https://github.com/utxostack/rgbpp-sdk/actions

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,13 +26,13 @@
     "@typescript-eslint/eslint-plugin": "^7.8.0",
     "@typescript-eslint/parser": "^7.8.0",
     "eslint": "^8.56.0",
-    "husky": "^9.0.11",
-    "lint-staged": "^15.2.2",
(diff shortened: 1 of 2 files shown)


ShookLyngs (approved):
I have a bit of concern about upgrading `tsup` since related issues have occurred before. Can you check if all packages work as expected after the snapshot is published?

duanyytop:
> I have a bit of concern about upgrading `tsup` since related issues have occurred before. Can you check if all packages work as expected after the snapshot is published?

I will check it later

duanyytop:
> > I have a bit of concern about upgrading `tsup` since related issues have occurred before. Can you check if all packages work as expected after the snapshot is published?
> 
> I will check it later

JoyID and [typeid-contract-cell-deps](https://github.com/utxostack/typeid-contract-cell-deps) work well.

PR #297, opened 2025-01-16
Add USDI to compatible xUDT list

duanyytop:
## Changes

- Add USDI to compatible xUDT list
- Fetch and cache compatible xUDT list from Vercel or GitHub server
- Use local static compatible xUDT list when the cache is empty
- Remove jsdelivr CDN because CDN cache time is too long, causing UTXO Airdrop cellDeps to become outdated
- Update UtxoAirdropBadge `cellDeps` test

### DIFF
--- a/examples/rgbpp/xudt/compatible-xudt/1-ckb-leap-btc.ts
+++ b/examples/rgbpp/xudt/compatible-xudt/1-ckb-leap-btc.ts
@@ -1,6 +1,6 @@
 import { serializeScript } from '@nervosnetwork/ckb-sdk-utils';
 import { genCkbJumpBtcVirtualTx } from 'rgbpp';
-import { getSecp256k1CellDep, buildRgbppLockArgs } from 'rgbpp/ckb';
+import { getSecp256k1CellDep, buildRgbppLockArgs, CompatibleXUDTRegistry } from 'rgbpp/ckb';
 import { CKB_PRIVATE_KEY, isMainnet, collector, ckbAddress, BTC_TESTNET_TYPE } from '../../env';
 
 interface LeapToBtcParams {
@@ -17,6 +17,13 @@ const leapRusdFromCkbToBtc = async ({
   compatibleXudtTypeScript,
 }: LeapToBtcParams) => {
   const toRgbppLockArgs = buildRgbppLockArgs(outIndex, btcTxId);
+
+  // Refresh the cache by fetching the latest compatible xUDT list from the specified URL.
+  // The default URL is:
+  // https://raw.githubusercontent.com/utxostack/typeid-contract-cell-deps/main/compatible-udt.json
+  // You can set your own trusted URL to fetch the compatible xUDT list.
+  // await CompatibleXUDTRegistry.refreshCache("https://your-own-trusted-compatible-xudt-url");
+  await CompatibleXUDTRegistry.refreshCache();
 
   const ckbRawTx = await genCkbJumpBtcVirtualTx({
     collector,
--- a/packages/ckb/src/constants/index.ts
+++ b/packages/ckb/src/constants/index.ts
@@ -66,15 +66,6 @@ const TestnetInfo = {
     depType: 'code',
   } as CKBComponents.CellDep,
 
-  CompatibleXUDTTypeScripts: [
-    // RUSD
-    {
-      codeHash: '0x1142755a044bf2ee358cba9f2da187ce928c91cd4dc8692ded0337efa677d21a',
-      hashType: 'type',
-      args: '',
-    },
-  ] as CKBComponents.Script[],
-
   UniqueTypeScript: {
     codeHash: '0x8e341bcfec6393dcd41e635733ff2dca00a6af546949f70c57a706c0f344df8b',
     hashType: 'type',
@@ -223,15 +214,6 @@ const MainnetInfo = {
     depType: 'code',
   } as CKBComponents.CellDep,
 
-  CompatibleXUDTTypeScripts: [
-    // RUSD
-    {
-      codeHash: '0x26a33e0815888a4a0614a0b7d09fa951e0993ff21e55905510104a0b1312032b',
-      hashType: 'type',
-      args: '',
-    },
-  ] as CKBComponents.Script[],
-
   UniqueTypeScript: {
     codeHash: '0x2c8c11c985da60b0a330c61a85507416d6382c130ba67f0c47ab071e00aec628',
     hashType: 'data1',
@@ -293,6 +275,33 @@ const MainnetInfo = {
     args: '',
   } as CKBComponents.Script,
 };
+
+export const COMPATIBLE_XUDT_TYPE_SCRIPTS: CKBComponents.Script[] = [
+  // RUSD Mainnet
+  {
+    codeHash: '0x26a33e0815888a4a0614a0b7d09fa951e0993ff21e55905510104a0b1312032b',
+    hashType: 'type',
+    args: '',
+  },
+  // RUSD Testnet
+  {
+    codeHash: '0x1142755a044bf2ee358cba9f2da187ce928c91cd4dc8692ded0337efa677d21a',
+    hashType: 'type',
+    args: '',
+  },
+  // USDI Mainnet
+  {
+    codeHash: '0xbfa35a9c38a676682b65ade8f02be164d48632281477e36f8dc2f41f79e56bfc',
+    hashType: 'type',
+    args: '',
+  },
+  // USDI Testnet
+  {
+    codeHash: '0xcc9dc33ef234e14bc788c43a4848556a5fb16401a04662fc55db9bb201987037',
+    hashType: 'type',
+    args: '',
+  },
+];
 
 export const UNLOCKABLE_LOCK_SCRIPT = {
   codeHash: '0x0000000000000000000000000000000000000000000000000000000000000000',
@@ -362,9 +371,6 @@ export const getSporeTypeScript = (isMainnet: boolean) =>
   isMainnet ? MainnetInfo.SporeTypeScript : TestnetInfo.SporeTypeScript;
 export const getSporeTypeDep = (isMainnet: boolean) =>
   isMainnet ? MainnetInfo.SporeTypeDep : TestnetInfo.SporeTypeDep;
-
-export const getCompatibleXudtTypeScripts = (isMainnet: boolean) =>
-  isMainnet ? MainnetInfo.CompatibleXUDTTypeScripts : TestnetInfo.CompatibleXUDTTypeScripts;
 
 export const getUtxoAirdropBadgeTypeScript = (isMainnet: boolean) =>
   isMainnet ? MainnetInfo.UtxoAirdropBadgeTypeScript : TestnetInfo.UtxoAirdropBadgeTypeScript;
--- a/packages/ckb/src/constants/index.ts
+++ b/packages/ckb/src/constants/index.ts
@@ -148,7 +148,7 @@ const TestnetInfo = {
   } as CKBComponents.Script,
 
   UtxoAirdropBadgeTypeDep: {
-    outPoint: { txHash: '0xfa0a6821293cc1ef4ee67a900862208e27f67b98237c9b13bf93c84607c5cd33', index: '0x2' },
+    outPoint: { txHash: '0xbbbb73972ac260a0f7204bea707288c3970688fe8714c3246a5e9a538168a42a', index: '0x0' },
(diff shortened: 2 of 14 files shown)


packages/ckb/src/constants/index.ts
```diff
@@ -294,6 +276,33 @@ const MainnetInfo = {
   } as CKBComponents.Script,
 };
 
+export const COMPATIBLE_XUDT_TYPE_SCRIPTS: CKBComponents.Script[] = [
```
Flouse, line 279: If `COMPATIBLE_XUDT_TYPE_SCRIPTS` is maintained statically in the rgbpp-sdk, it will come out when a new xUDT compatible token is released, all dApps using rgbpp-sdk will need to update rgbpp-sdk.

packages/ckb/src/utils/ckb-tx.ts
```diff
@@ -44,8 +44,8 @@ export const isTokenMetadataType = (type: CKBComponents.Script, isMainnet: boole
   return tokenMetadataType === typeAsset;
 };
 
-export const isCompatibleUDTTypesSupported = (type: CKBComponents.Script, isMainnet: boolean): boolean => {
-  const compatibleXudtTypeBytes = getCompatibleXudtTypeScripts(isMainnet).map((script) => serializeScript(script));
+export const isCompatibleUDTTypesSupported = (type: CKBComponents.Script): boolean => {
+  const compatibleXudtTypeBytes = COMPATIBLE_XUDT_TYPE_SCRIPTS.map((script) => serializeScript(script));
```
Flouse, line 48: Maybe the compatible xUDT token list could be fetched from a trusted API endpoint, such as:
- https://raw.githubusercontent.com/utxostack/typeid-contract-cell-deps/main/compatible-udt.json
- or a Vercel deployment by utxostack/typeid-contract-cell-deps


and then cache it locally for 3 mins ([about 24 CKB blocks](https://docs-old.nervos.org/docs/essays/tx-confirmation))

## draft

```draft code
class CompatibleXUDTRegistry {
  private static cache: CKBComponents.Script[] | null = null
  private static lastFetchTime: number = 0
  private static CACHE_DURATION = 3 * 60 * 1000 // 3 minutes ([about 24 CKB blocks](https://docs-old.nervos.org/docs/essays/tx-confirmation))
  
  static getCompatibleTokens(): CKBComponents.Script[] {
    const now = Date.now()
    if (!this.cache || now - this.lastFetchTime > this.CACHE_DURATION) {
      this.refreshCache()
    }
    return this.cache || COMPATIBLE_XUDT_TYPE_SCRIPTS
  }

  private static async refreshCache(): Promise<void> {
    try {
      const response = await fetch('https://raw.githubusercontent.com/utxostack/typeid-contract-cell-deps/main/compatible-udt.json') // TODO: race fetch from vercel deployment
      this.cache = await response.json()
      this.lastFetchTime = Date.now()
    } catch {
      // Fallback to default list if fetch fails
      
      this.cache = COMPATIBLE_XUDT_TYPE_SCRIPTS
    }
  }
}
```

packages/ckb/src/constants/index.ts
duanyytop: I tried to fetch compatible xUDT list from GitHub of Vercel server in https://github.com/utxostack/rgbpp-sdk/pull/296.

However, these changes will cause incompatible upgrades. Judging from the current situation of ecological projects, maintaining a static list is an acceptable solution.

ShookLyngs: Is it acceptable to add an async version of the `isUDTTypeSupported` method?

duanyytop: @ShookLyngs This is also the issue I want to discuss.

duanyytop: Then we need to maintain two functions. Which function do we recommend third-party developers use?

packages/ckb/src/utils/ckb-tx.ts
ShookLyngs: One issue with the `getCompatibleTokens()` function is that it doesn't always (or guarantee to) return the latest data. In worse cases, the caller might be forced to request the API multiple times to verify the information.

packages/ckb/src/constants/index.ts
ShookLyngs: Or:
```
isUDTTypeSupported(..., fetchLatest: false): boolean | Promise<boolean>;
isUDTTypeSupported(..., fetchLatest: true): Promise<boolean>;
```

ShookLyngs: Another idea, **for anyone who needs the tip information**, they should fetch the latest version of the scripts before calling any rgbpp-sdk/ckb functions. For example, they can call a function like this:

```ts
await fetchLatestXXXScript(); // call this
await initCkbScripts(); // or this
```

duanyytop: > Another idea, **for anyone who needs the tip information**, they should fetch the latest version of the scripts before calling any rgbpp-sdk/ckb functions. For example, they can call a function like this:
> 
> ```ts
> await fetchLatestXXXScripts(network); // call this
> await initCkbScripts(network); // or this
> ```

Is it another big breaking change?

packages/ckb/src/utils/ckb-tx.ts
duanyytop: Yes. 

Besides, the cache only stays in the memory of users' devices, if they restart Dapp, the cache will be cleared and must be fetched again.

Flouse: > One issue with the `getCompatibleTokens()` function is that it doesn't always (or guarantee to) return the latest data. In worse cases, the caller might be forced to request the API multiple times to verify the information.


To mitigate the problem:
let‘s say there is a rgbpp-sdk init step, and `CompatibleXUDTRegistry.refreshCache` could be called in that step.

packages/ckb/src/constants/index.ts
ShookLyngs: > Is it another big breaking change?

I think this is not a breaking change, because not everyone should use it.
Only if the caller strically need the latest version of the scripts, you should use the new function.

examples/rgbpp/xudt/compatible-xudt/1-ckb-leap-btc.ts
```diff
@@ -18,6 +18,13 @@ const leapRusdFromCkbToBtc = async ({
 }: LeapToBtcParams) => {
   const toRgbppLockArgs = buildRgbppLockArgs(outIndex, btcTxId);
 
+  // Refresh the cache by fetching the latest compatible xUDT list from the specified URL.
+  // The default URL is:
+  // https://raw.githubusercontent.com/utxostack/typeid-contract-cell-deps/main/compatible-udt.json
+  // You can set your own trusted URL to fetch the compatible xUDT list.
+  // await CompatibleXUDTRegistry.refreshCache("https://your-own-trusted-compatible-xudt-url");
+  await CompatibleXUDTRegistry.refreshCache();
```
Flouse, line 26: 👍

packages/ckb/src/constants/index.ts
```diff
@@ -294,6 +276,33 @@ const MainnetInfo = {
...
+    codeHash: '0xbfa35a9c38a676682b65ade8f02be164d48632281477e36f8dc2f41f79e56bfc',
+    hashType: 'type',
+    args: '',
+  },
+  // USDI Testnet
+  {
+    codeHash: '0xcc9dc33ef234e14bc788c43a4848556a5fb16401a04662fc55db9bb201987037',
+    hashType: 'type',
+    args: '',
+  },
```
Flouse, line 303: https://github.com/utxostack/typeid-contract-cell-deps/blob/main/compatible-udt.json does not yet include USDI.

Are you going to update it later and verify the process designed after this PR?
Or what is the plan?

duanyytop: I am adding USDI to [compatible-udt.json](https://github.com/utxostack/typeid-contract-cell-deps/blob/main/compatible-udt.json) now.

packages/ckb/src/utils/cell-dep.ts
```diff
@@ -210,3 +202,58 @@ export const fetchTypeIdCellDeps = async (
...
+export class CompatibleXUDTRegistry {
+  private static cache: CKBComponents.Script[] = [];
+  private static lastFetchTime: number = 0;
+  private static CACHE_DURATION = 3 * 60 * 1000; // 3 minutes ([about 24 CKB blocks](https://docs-old.nervos.org/docs/essays/tx-confirmation))
+  private static xudtUrl = VERCEL_STATIC_COMPATIBLE_XUDT_URL;
+
+  // If you want to get the latest compatible xUDT list, CompatibleXUDTRegistry.refreshCache should be called first
+  static getCompatibleTokens(): CKBComponents.Script[] {
+    const now = Date.now();
+    if (this.cache.length > 0 || now - this.lastFetchTime > this.CACHE_DURATION) {
```
Flouse, line 223: The condition `this.cache.length > 0` will refresh the cache every time.


```suggestion
    if (this.cache.length === 0 || now - this.lastFetchTime > this.CACHE_DURATION) {
```

packages/ckb/src/constants/index.ts
duanyytop: https://github.com/utxostack/typeid-contract-cell-deps/pull/19

packages/ckb/src/utils/cell-dep.ts
```diff
@@ -210,3 +202,58 @@ export const fetchTypeIdCellDeps = async (
...
+   * fetch time to the current timestamp.
+   *
+   * @param url - An optional URL to fetch data from. If not provided, a default
+   * URL (`VERCEL_CELL_DEPS_JSON_STATIC_URL`) will be used.
+   * @returns A promise that resolves when the cache has been refreshed.
+   */
+  static async refreshCache(url?: string): Promise<void> {
+    this.xudtUrl = url ?? VERCEL_CELL_DEPS_JSON_STATIC_URL;
+    try {
+      const response = await Promise.any([request(this.xudtUrl), request(GITHUB_STATIC_COMPATIBLE_XUDT_URL)]);
```
Flouse, line 244: When developers explicitly provide a URL, we should respect their choice and not fall back to the GitHub URL `GITHUB_STATIC_COMPATIBLE_XUDT_URL`.

Flouse (changes requested):

packages/ckb/src/utils/cell-dep.ts
duanyytop: Updated in https://github.com/utxostack/rgbpp-sdk/pull/297/commits/47c7e8d642892c267c15f8ff122e5e0e010c5e30

packages/ckb/src/utils/cell-dep.ts
duanyytop: Fixed in https://github.com/utxostack/rgbpp-sdk/pull/297/commits/47c7e8d642892c267c15f8ff122e5e0e010c5e30

packages/ckb/src/constants/index.ts
duanyytop: Tested with USDI in https://github.com/utxostack/rgbpp-sdk/pull/297/commits/47c7e8d642892c267c15f8ff122e5e0e010c5e30

packages/ckb/src/utils/cell-dep.ts
```diff
@@ -210,3 +202,61 @@ export const fetchTypeIdCellDeps = async (
...
+   * successful, it updates the cache with the fetched data and sets the last
+   * fetch time to the current timestamp.
+   *
+   * @param url - An optional URL to fetch data from. If not provided, a default
+   * URL (`VERCEL_CELL_DEPS_JSON_STATIC_URL`) will be used.
+   * @returns A promise that resolves when the cache has been refreshed.
+   */
+  static async refreshCache(url?: string): Promise<void> {
+    this.xudtUrl = url ?? VERCEL_CELL_DEPS_JSON_STATIC_URL;
+    const isExternal = url !== VERCEL_CELL_DEPS_JSON_STATIC_URL && url !== GITHUB_STATIC_COMPATIBLE_XUDT_URL;
```
Flouse, line 243: VERCEL_CELL_DEPS_JSON_STATIC_URL -> VERCEL_STATIC_COMPATIBLE_XUDT_URL

Flouse (changes requested):

Flouse (approved):

ShookLyngs (approved):

PR #313, opened 2025-02-26
fix: handle edge case when change amount is below minimum capacity

fghdotio:
When `actualInputsCapacity > sumOutputsCapacity` but the resulting change amount  is below `MIN_CAPACITY`, we also need to collect additional cells to ensure the change  output meets the minimum capacity requirement. This prevents creation of invalid  cells with insufficient capacity.

### DIFF
--- a/packages/ckb/src/rgbpp/btc-transfer.ts
+++ b/packages/ckb/src/rgbpp/btc-transfer.ts
@@ -382,18 +382,30 @@ export const appendIssuerCellToBtcBatchTransferToSign = async ({
 
   let actualInputsCapacity = BigInt(sumInputsCapacity);
   const txFee = MAX_FEE;
-  if (actualInputsCapacity <= sumOutputsCapacity) {
+  let changeCapacity = actualInputsCapacity - sumOutputsCapacity;
+  while (changeCapacity < MIN_CAPACITY) {
     const needCapacity = sumOutputsCapacity - actualInputsCapacity + MIN_CAPACITY;
     const { inputs, sumInputsCapacity: sumEmptyCapacity } = collector.collectInputs(emptyCells, needCapacity, txFee);
     rawTx.inputs = [...rawTx.inputs, ...inputs];
     actualInputsCapacity += sumEmptyCapacity;
+    changeCapacity = actualInputsCapacity - sumOutputsCapacity;
   }
 
-  let changeCapacity = actualInputsCapacity - sumOutputsCapacity;
   const changeOutput = {
     lock: issuerLock,
     capacity: append0x(changeCapacity.toString(16)),
   };
+  const secp256k1CellDep = getSecp256k1CellDep(isMainnet);
+  if (
+    !rawTx.cellDeps.some(
+      ({ outPoint }) =>
+        outPoint &&
+        outPoint.txHash === secp256k1CellDep.outPoint!.txHash &&
+        outPoint.index === secp256k1CellDep.outPoint!.index,
+    )
+  ) {
+    rawTx.cellDeps = [...rawTx.cellDeps, secp256k1CellDep];
+  }
   rawTx.outputs = [...rawTx.outputs, changeOutput];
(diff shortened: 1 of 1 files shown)


packages/ckb/src/rgbpp/btc-transfer.ts
```diff
@@ -382,18 +382,30 @@ export const appendIssuerCellToBtcBatchTransferToSign = async ({
 
   let actualInputsCapacity = BigInt(sumInputsCapacity);
   const txFee = MAX_FEE;
-  if (actualInputsCapacity <= sumOutputsCapacity) {
+  let changeCapacity = actualInputsCapacity - sumOutputsCapacity;
+  while (changeCapacity < MIN_CAPACITY) {
```
duanyytop, line 386: Why do you use a `while` loop? How do you return from the loop?

duanyytop: I think if the `changeCapacity < MIN_CAPACITY` is true, throwing an error is enough for issuers.

packages/ckb/src/rgbpp/btc-transfer.ts
```diff
@@ -382,18 +382,30 @@ export const appendIssuerCellToBtcBatchTransferToSign = async ({
...
     actualInputsCapacity += sumEmptyCapacity;
+    changeCapacity = actualInputsCapacity - sumOutputsCapacity;
   }
 
-  let changeCapacity = actualInputsCapacity - sumOutputsCapacity;
   const changeOutput = {
     lock: issuerLock,
     capacity: append0x(changeCapacity.toString(16)),
   };
+  const secp256k1CellDep = getSecp256k1CellDep(isMainnet);
```
duanyytop, line 398: The secp256k1 cell dep has been added in the function `genBtcBatchTransferCkbVirtualTx`, so why do you add secp256k1 cell dep again?

packages/ckb/src/rgbpp/btc-transfer.ts
fghdotio: Initially I mistook `txFee` for fee rate - I thought the loop ensures sufficient capacity to cover increasing transaction fees as the transaction size grows with each new input from `collectInputs`. As `actualInputsCapacity` accumulates in the loop, `changeCapacity` gradually increases until it either exceeds `MIN_CAPACITY` for normal exit, or throws `CapacityNotEnoughError` when available cells are insufficient.

Upon closer inspection after reading your comments, I realize a `while` loop is unnecessary since `txFee` is set to a sufficiently large `MAX_FEE` constant. A simple `if` statement would be adequate here.

Despite its name, `appendIssuerCellToBtcBatchTransferToSign` serves as a general fee supplement method in offline mode. Perhaps we should collect additional issuer's cells as inputs to cover tx fee whenever possible, instead of throwing an error. Please let me know your thoughts.

packages/ckb/src/rgbpp/btc-transfer.ts
fghdotio: `appendIssuerCellToBtcBatchTransferToSign` serves as a general fee supplement method in offline mode. While `genBtcBatchTransferCkbVirtualTx` does include the secp256k1 cell dep, other transaction builders like `genBtcTransferCkbVirtualTx` may not include it by default. That's why there's a conditional check to add the secp256k1 cell dep if it's not already present.

duanyytop: I see.

packages/ckb/src/rgbpp/btc-transfer.ts
duanyytop: 1. The `MAX_FEE` is 0.2CKB, and it is enough for almost transactions.
2. If the issuer's balance is not enough for collecting, the loop does not work.
3. I think you could provide a capacity estimate function for the issuer before collecting cells.

fghdotio: Good point! I'll try to implement it later.

